PR #9, opened 2018-10-02
Correct for unequal class weights inclass fractions of leaf populations

dschoenleber:
In the current implementation, class weights -- e.g. DecisionTreeClassifier(class_weight='balanced') -- are not corrected for when plotting the pie charts with the class fractions of the final leaf populations. This pull requests fixes this by 
* tracking the decision tree class weights in the shadow tree
* dividing by the class weights before plotting (if class weights is other than None)

### DIFF
--- a/dtreeviz/shadow.py
+++ b/dtreeviz/shadow.py
@@ -47,7 +47,6 @@ class ShadowDecTree:
         self.feature_names = feature_names
         self.class_names = class_names
         self.class_weight = tree_model.class_weight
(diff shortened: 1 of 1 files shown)


parrt:
Thanks very much! Can you reply with "I agree" to indicate you release your changes under MIT License?

dschoenleber:
I agree!
I pushed a little rearrangement to be more in line with the existing code.

parrt:
thanks!

Issue #20, opened 2018-11-14
Garbled characters occur when using Japanese because fontname is hard coded "Arial"


h4nyu:
Can I set fontname(if yes, how can I set) or let matplotlib.rcParams to set fontname?

https://github.com/parrt/dtreeviz/blob/09053833b46a92d75dbe0a9b98d16f35fb5f3394/dtreeviz/trees.py#L473-L474

![japanese_font](https://user-images.githubusercontent.com/18400546/48455739-b0948400-e7ff-11e8-9e82-e8b0f8e611e5.PNG)

parrt:
Hi. unfortunately, all I have is ability to set font size.  I would consider a pull request to fix.

parrt:
Fixed by https://github.com/parrt/dtreeviz/pull/24

PR #24, opened 2018-11-29
remove hard coded fontname

h4nyu:
#20


parrt:
hiya. seems better if we allow a parameter on that function, right?

PR #36, opened 2019-04-03
Remove references to svg:cairo.

gautamkarnik:
Changed the logic in trees.py and updated the README.md file as per my installation on Mac 10.14 and Windows 10.


parrt:
Only issue to think about is the font metrics problem when building HTML tables as part of tree construction.

parrt:
Works on 10.13 mac too. Will try with fewer lib installs.

PR #38, opened 2019-05-28
fixed ShadowDecTree missing parameter

H4dr1en:
This is a fix for the following error:

image


parrt:
Hi. thanks. that's actually dead code i should remove.

PR #39, opened 2019-05-29
Added colorbar to rtreeviz_bivar_heatmap

H4dr1en:
In bivariate feature-target space plots it is easy to understand what are the targeted values because they are represented as a dimension of the plot:
image

But in bivariate feature-target space heatmap, there is no indication at all of the values represented by the colors: 
image

This PR adds a colorbar to the heatmap so that it is easier to understand how colors are related to actual values:
image


parrt:
Thanks.  looks pretty cool. I will look at this hopefully in a month or so; brain stuck in something else.

PR #42, opened 2019-06-14
removed unused imports/local variables, clean-up of import/docstrings

Ashafix:
* Removed unused imports
* Removed unused local variables
* `import *` was replaced with the specific functions/objects
* doc strings were cleaned up


parrt:
Nice work. thanks!

PR #52, opened 2019-09-27
Add visualisations for leaf samples

tlapusan:
- vizualisations for leaf samples (both classification and regression)
- create a new notebook for them


parrt:
Looking good, @tlapusan !  In my routines, I immediately create a shadow tree to make my life easier.  Your current code is self-contained but I wonder if we should be consistent somehow?  Particularly if there are other PRs coming. Maybe we can move your utility code like `_get_leaf_samples()` into https://github.com/parrt/dtreeviz/blob/master/dtreeviz/shadow.py ?

tlapusan:
Hi @parrt, I looked over ShadowDecTree and it seems the right place to add the utility methods. But, the ShadowDecTree constructor contains a lot of logic and requires mandatory arguments like X_train, y_train, feature_names, even if I don't need them to render sample leaves(the same case for other visualisations I will plan to add).  I added these arguments to see how the code looks now.

Do you have any suggestions how we could make ShadowDecTree constructor as light as possible and to call all its other functionalities only when it's needed ?  Thanks.

parrt:
What if we made your new code static functions so we'd use ShadowDecTree.foo() but wouldn't need ShadowDecTree constructor to use?

tlapusan:
> What if we made your new code static functions so we'd use ShadowDecTree.foo() but wouldn't need ShadowDecTree constructor to use?

It worked ;)

parrt:
Looking great. A minor function name and let's merge!!! thanks for the great contribution.

parrt:
I can tweak after merge if you want.

parrt:
Oh and I might wanna change the look once I merge and see it. Wanna make sure it's consistent with, for example: https://github.com/parrt/dtreeviz/blob/master/testing/samples/iris-TD-2.svg

tlapusan:
sure, you can merge and make the final tweaks. I tried to choose the right colors from colors.COLORS, but you know better what colors make the fit. Meanwhile, I will start working to add another type of visualisation, for example to visualise the leaf criterion (gini, entropy for classification and MAE, MSE for regression)

parrt:
Ok, I think it's more useful to show the actual prediction in the leaves so maybe that can be added to the plot. not sure how useful the entropy or MSE is; at least I've never used it for anything.

Can you make a PR that makes the histograms look the same? At least in the previous pictures the bars were not connected like they are in my plots; is not just a color thing. thanks for the hard work!

PR #55, opened 2019-10-11
Adjust histogram look

tlapusan:
- remove the space between bars
- related to the hist colors : 
@parrt, you made a reference to this picture : https://github.com/tlapusan/dtreeviz/blob/master/testing/samples/iris-TD-2.svg. This is an histogram which adapts its colors based on number of classes. In ctreeviz_leaf_samples picture, we have a more general histogram, showing the distribution of leaf samples (not related to classes). 
I found in colors.py the LIGHTBLUE constant and assign this color to the ctreeviz_leaf_samples histogram, but I'm not sure if it's the right one. If not, please suggest what color would you prefer :)


parrt:
Cool. i'll merge and take a look. Less  concerned with the color and more about the look of the histograms.

PR #57, opened 2019-10-12
Remove ax spines (top and right), outlined bars

### LINKED ISSUE #56: outlines in leaf bar charts
hi @tlapusan,  The leaf graphs are looking closer to the other style now but I think I'd prefer if you remove the background grid as it's different. If you still wanted, you could put an option as a parameter I guess.

 Also, note that I have an outline around the bars as a looks a lot better:

Screen Shot 2019-10-11 at 11 05 52 AM

See those grey outlines?  I think the code that does that is (such as in `class_split_viz()`):

```
        hist, bins, barcontainers = ax.hist(X_hist,
                                            color=X_colors,
                                            align='mid',
                                            histtype=histtype,
                                            bins=bins,
                                            label=class_names)
        # Alter appearance of each bar
        for patch in barcontainers:
            for rect in patch.patches:
                rect.set_linewidth(.5)
                rect.set_edgecolor(colors['rect_edge'])
```

tlapusan:
leaf_samples

I made some adjustments to the look of ctreeviz_leaf_samples(). It looks pretty cool now :)
If you think that everything is fine, I will add all the adjustment to the other visualisation, viz_leaf_samples_by_class()

close #56


parrt:
Looks much better! Please do add to other visualizations.

PR #58, opened 2019-10-16
adjust look for viz_leaf_samples_by_class 

tlapusan:
hi @parrt,
I adjusted the look for the viz_leaf_samples_by_class().

Is there a reason why you chose the previous method name to ctreeviz_leaf_samples()? Should I change the name of viz_leaf_samples_by_class to ctreeviz_leaf_samples_by_class ?


tlapusan:
This is how it looks now.

Screenshot 2019-10-16 at 17 15 02

parrt:
Dang. I messed up. You're right. I'll change back after merge. I meant to do ctreeviz_leaf_samples for viz_leaf_samples_by_class, not for the other function. ctreeviz_leaf_samples is consistent with the other names and ctree implies leaf samples would be shown per class.

parrt:
Fixed names in https://github.com/parrt/dtreeviz/commit/aca1aed0611b2029ff9498fed6de4481d6457593

PR #59, opened 2019-10-18
Fix examples to use dtreeviz #54

### LINKED ISSUE #54: Fix examples to use dtreeviz
In [new examples](https://github.com/parrt/dtreeviz/blob/master/notebooks/tree_structure_example.ipynb)  the trees are displayed using the default sklearn not dtreeviz ;)  This is for @tlapusan :)

tlapusan:
Display the tree graph using dtreeviz instead of standard sklearn visualisation.

Close #54


parrt:
thanks!

PR #62, opened 2019-10-20
fix : the right arrows in path to test data are not highlighted

kento-forest:
In the current implementation, the right arrows in path to test data are not highlighted. For example, 

<div align="center">
before
</div>

This pull request fixes this by modifying rcolor in dtreeviz/tree.py
After modifying,

<div align="center">
after
</div>


parrt:
thanks, good catch!

PR #64, opened 2019-11-02
Create the visualisation method for leaf target distribution

### LINKED ISSUE #63: Make leaf split plot for regressors to show distribution of leaf sample y's
I wonder if a split plot per leaf would look ok; seems like it'd be useful, @tlapusan.  Here is a sample split plot.  Maybe one y range / axis on left and then tight grouping of split plots?

Screen Shot 2019-10-28 at 10 44 58 AM

tlapusan:
Create the first version for this visualisation. 
We could add new improvements on this PR if it's needed.

Close #63

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -1416,8 +1416,7 @@ def ctreeviz_leaf_samples(tree_model: tree.DecisionTreeClassifier,
 def _get_leaf_target_input(shadow_tree: ShadowDecTree,
                            y_train,
                            target_name: str,
-                           precision: int,
-                           leaf_space: int):
+                           precision: int):
     x = []
     y = []
     means = []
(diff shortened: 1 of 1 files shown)


parrt:
From #63,

parrt:
Hmm...here's a thought...what if you displayed the leaf id as y axis not horizontal axis?  Then large trees get taller, not wider graphs and we don't need the splitting of graphs?

tlapusan:
Very nice idea! Will try to implement it

parrt:
Maybe we should do the same for your other leaf visualizations?

tlapusan:
> Maybe we should do the same for your other leaf visualizations?

yes, it could be a valid option. Let's see how it looks for this visualisation and we will decide afterward.

parrt:
can you send a sample pic of horizontal?

PR #71, opened 2019-12-01
Show just path prediction

### LINKED ISSUE #70: show_just_path_prediction
I started to work on showing just the prediction path (based on what we discussed in a previous issue)

Right now, it looks like this : 
Screenshot 2019-11-28 at 15 01 21

Am I on the right track ? :) Do you think we need to show also the neighborhood nodes. I don't have a strong argument for them, but I like how they look (they show somehow the 'opposite' prediction path)

tlapusan:
PR based on issue #70
close #70


parrt:
Nice. i like the way you factored that common code out!

PR #77, opened 2019-12-29
Create specialized method for node samples' description

tlapusan:
method call example : `describe_node_sample(dtc, dataset[features], node_id=58)`


parrt:
thanks!

PR #80, opened 2020-02-25
Feature importance instance prediciton path #79

### LINKED ISSUE #79: Feature importance for instance prediction
Hi @parrt,
In another issue (#65 ), we have discussed about making a prediction more English readable. One solution was to show the feature importances only for nodes on the prediction path. 
I have looked into sklearn source code and I have found how feature importance is calculated for a decision tree. I could extract that code and adapt it to get the feature importance for a prediction path. 
I could come up with a visualisation sample in the next days.
What do you think ? :)

tlapusan:
close #79


parrt:
I'm not sure we should introduce a new directory. Can we simply make a new file in the same dtreeviz src dir or perhaps even incorporated into one of the other files so we don't need another import when using?

tlapusan:
Let's see if I understand correctly... explain_prediction_path() method is implemented in dtreeviz module, so no need to import any other module. Dtreeviz and the other modules can become pretty big in the future...that's why I wanted to package/modularize next big functionalities. This new package contains the business logic for prediction path explanation... We can think ro a new approach if you consider it

parrt:
What if we simply made a new file called interpretation.py? That way we don't introduce packages, just another module.

tlapusan:
It sounds good a new interpretation.py module. Right now I'm in a ski trip, hope to have find a window to refractor it.

tlapusan:
@parrt I made the refactoring.

parrt:
something big happened globally to formatting. did you run a formatter by chance? Makes it hard to see diffs ;)

tlapusan:
I'm using pycharm for development with its default formatter. What IDE are you using ? I agree that we should be using the same formatter, to eliminate these scenarios. 

Most of the differences are caused by spacing between operators. PEP8 recommends using a single space on either side of operators (https://www.python.org/dev/peps/pep-0008/#other-recommendations) and I think it should be a good practice to follow PEP8 guidance.  

If you have other preferences, I could use your formatter.

parrt:
Well, I almost never rely on a formatter except locally.  If you are formatting over the entire file, maybe that's not a good idea.  at least some of the changes were bad in my opinion ;)   can you undo that? I also use PyCharm btw :)

tlapusan:
Sure, it should be possible to undo it. Is an old habit to use shortcuts for autoformatter.

parrt:
:D hahaha.  it's still not possible to see what the changes really are because of all of the white space changes.  any chance you could simply reapply your updates to the original file? Maybe we need a new PR?  It sounds like there's just some additions to one file and then a new file to add so shouldn't be too much work. sorry, just trying to be clean. :)

tlapusan:
In the last commit, I compared my trees.py branch version with master version and applied all code style differences.
If I'm looking in tab "Files changes" from this PR, only the new added method appears as a difference (https://github.com/parrt/dtreeviz/pull/80/files) : 

Screenshot 2020-03-04 at 21 25 49

btw, what tool are you using to see the differences in this PR ?

parrt:
Wow. it's as you say now. maybe I looked at an old commit? anyway, looks great. thanks for the hard work!

PR #82, opened 2020-03-28
dtreeviz/utils.py: Fix ID collisions in inline_svg_images output

CyberShadow:
Inlined SVGs may have IDs which are not globally unique. Patch the tree of the inlined images and add a prefix to all IDs, to make collisions much less likely.

Please excuse my Python.

### DIFF
--- a/dtreeviz/utils.py
+++ b/dtreeviz/utils.py
@@ -45,7 +45,7 @@ def inline_svg_images(svg) -> str:
 
     # Find all image tags in document (must use svg namespace)
     image_tags = tree.findall(".//svg:g/svg:image", ns)
-    for img in image_tags:
+    for i, img in enumerate(image_tags):
         # load ref'd image and get svg root
         svgfilename = img.attrib["{http://www.w3.org/1999/xlink}href"]
         with open(svgfilename, encoding='UTF-8') as f:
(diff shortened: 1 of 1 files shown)


parrt:
Interesting! Can you tell me what problems occurred because of this? In other words how did you know this was a problem? thanks!

CyberShadow:
I found a link to this file in a StackOverflow post and applied it to my toolchain for putting LaTeX formulas in Graphviz SVG renderers... sorry if this is not useful to the project as a whole :)

parrt:
I've noticed that jupyter notebooks tend not to show my SVG images except the first time you run it, not when you reload it. You think that could be because of this unique identifier thing? Can you give me a link to the StackOverflow post? I want to understand exactly what's going on. thanks

CyberShadow:
> I've noticed that jupyter notebooks tend not to show my SVG images except the first time you run it, not when you reload it. You think that could be because of this unique identifier thing?

Hmm, I don't think so. The symptoms for ID collisions (which I also encountered elsewhere) is things such as styling applied to the wrong element. In this case it manifested as a wrong font in some of the SVGs embedded later in the document.

> Can you give me a link to the StackOverflow post? I want to understand exactly what's going on. thanks

Sorry, I misremembered. Though I did run into dtreeviz on StackOverflow while researching my SVG troubles, the actual link to this file I found on this GitHub comment:
https://github.com/ellson/MOTHBALLED-graphviz/issues/1294#issuecomment-425730041

PR #84, opened 2020-04-01
Viz leaf split criterion

### LINKED ISSUE #81: Visualize leaf criterion
hi @parrt,

leaves from regressor and classifier trees contains two important information : number of samples and criterion.
Criterion for regressor are “mse”, “friedman_mse”, “mae” and for classifer are "gini" and "entropy". All of them shows the leaf performance/confidence for its predictions.

I think would be helpful to have some visualisations for leaf criterion. Each leaf performance, in the end, will determine the general tree performance.
It's also useful to interpret leaf criterion in parallel with leaf samples. For ex. a leaf could have a very good performance (ex. gini -> 0), but that leaf contains very few samples... so that leaf will overfit. The best scenario for a leaf is to have good performance and many leaf samples.

Screen Shot 2020-03-23 at 5 36 07 PM

We can also plot the histogram of leaf criterion. This is useful when the tree has a lot of leaves or when you want to have a general point-of-view about all leaf criterion (general tree performance).

Screen Shot 2020-03-23 at 5 41 36 PM

I already have an implementation for these visualisations (95% ready for a PR).
What do you think about it ? :)

tlapusan:
PR based on #81 task.

close #81

### DIFF
--- a/dtreeviz/shadow.py
+++ b/dtreeviz/shadow.py
@@ -253,6 +253,7 @@ class ShadowDecTree:
         x, y = zip(*leaf_samples)
         return np.array(x), np.array(y)
 
+    @staticmethod
     def get_leaf_criterion(_tree_model):
         """Get criterion for each leaf
         For classification, supported criteria are “gini” for the Gini impurity and “entropy” for the information gain.
--- a/dtreeviz/shadow.py
+++ b/dtreeviz/shadow.py
@@ -253,6 +253,19 @@ class ShadowDecTree:
         x, y = zip(*leaf_samples)
         return np.array(x), np.array(y)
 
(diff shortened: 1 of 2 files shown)


parrt:
looks good. btw, should it be `loss` or `criterion`?

tlapusan:
Criterion is the common attribute in sklearn for both DecisionTreeClassifier and DecisionTreeRegressor to measure the quality of a node...
Yes...criterion doesn't sounds very good for this method...I guess...but didn't have a better one during development :) viz_leaf_performance ? viz_leaf_quality ?

parrt:
we should be consistent with sklearn. criterion it is then!

PR #105, opened 2020-09-29
Update readme

tlapusan:
hi @parrt 

I added few more visualizations into readme page and other few modifications. Please take a look and adjust if you consider.


parrt:
Looks great. thanks for fixing!

PR #112, opened 2020-11-27
Support lightgbm decision trees

tlapusan:


parrt:
It looks like this is a combined categorical split and lightgbm thing.  I really think we should separate those two.  We can do the lightgbm merge before we can figure out the categorical split thing I would say, right?

parrt:
also it looks like you will have to re-base.  Note that I have changed the testing/dtreeviz directory name to avoid  module name issues.

setup.py
```diff
@@ -8,7 +8,8 @@
     url='https://github.com/parrt/dtreeviz',
     license='MIT',
     packages=find_packages(),
-    install_requires=['graphviz>=0.9','pandas','numpy','scikit-learn','matplotlib','colour', 'xgboost', 'pytest', 'pyspark'],
+    install_requires=['graphviz>=0.9','pandas','numpy','scikit-learn','matplotlib','colour', 'xgboost', 'pytest', 'pyspark',
```
parrt, line 11: during your cleanup, don't forget to re-base to pull in the dynamic loading of libraries :) I think we merged a PR for that, but I could be confusing it with one of the other billion Python libraries I'm trying to track in my head right now. haha!

tlapusan: Sure, I've already merged upstream master with my local lightgbm branch

parrt: Ah right. ok, i'll hold off on this PR until you've updated. thanks!

dtreeviz/models/shadow_decision_tree.py
```diff
@@ -240,14 +245,17 @@ def get_score(self) -> float:
...
         pass
 
     @abstractmethod
     def shouldGoLeftAtSplit(self, id, x):
         """Return true if it should go to the left node child based on node split criterion and x value"""
         pass
 
+    def get_root_node_labels(self):
+        pass
+
```
parrt, line 258: What's a root note label? Do you mean edges?

dtreeviz/models/sklearn_decision_trees.py
```diff
@@ -15,6 +15,7 @@ def __init__(self, tree_model,
                  target_name: str = None,
                  class_names: (List[str], Mapping[int, str]) = None):
 
+        self.tree_model = tree_model
```
parrt, line 18: why set model here when super init does it?

dtreeviz/models/sklearn_decision_trees.py
```diff
@@ -105,13 +119,13 @@ def get_feature_path_importance(self, node_list):
 
                 gini_importance[self.tree_model.tree_.feature[node]] += self.tree_model.tree_.weighted_n_node_samples[
                                                                             node] * \
-                                                                        self.tree_model.tree_.impurity[node] \
-                                                                        - self.tree_model.tree_.weighted_n_node_samples[
+                                                                     self.tree_model.tree_.impurity[node] \
```
parrt, line 122: formatting is all shifted too far to the right

setup.py
```diff
@@ -10,7 +10,7 @@
     packages=find_packages(),
     install_requires=['graphviz>=0.9','pandas','numpy','scikit-learn',
                         'matplotlib','colour', 'pytest'],
-    extras_require={'xgboost': ['xgboost'], 'pyspark':['pyspark']},
+    extras_require={'xgboost': ['xgboost'], 'pyspark':['pyspark'], 'lightgbm':['lightgbm']},
```
parrt, line 13: might need to tweak readme install doc

dtreeviz/models/shadow_decision_tree.py
```diff
@@ -284,7 +292,7 @@ def get_split_node_heights(self, X_train, y_train, nbins) -> Mapping[int, int]:
             # print(f"\tmax={np.max(height_of_bins):2.0f}, heights={list(height_of_bins)}, {len(height_of_bins)} bins")
         return node_heights
 
-    def predict(self, x: np.ndarray) -> Tuple[Number, List]:
+    def predict(self, x: np.ndarray, path_only: bool = False) -> Tuple[Number, List]:
```
parrt, line 295: why add new arg? For use, `path = self.predict(self.x_data[i], path_only=True)` just do `_, path = self.predict(self.x_data[i])`

tlapusan: if I don't send path_only arg, we will receive "RecursionError: maximum recursion depth exceeded while decoding a JSON array from a unicode string" for LightGBM.
This is because get_node_samples() call the predict() method and leaf.prediction() (from inside predict()) will also call for get_node_samples to calculate the leaf prediction.

tlapusan:
I see that we have some conflicts from notebooks....hard to merge on them. What do you suggest to do with them ? 
Since the notebooks are not related to lightgbm, maybe we should accept the notebooks from master.

parrt:
Yep, i think that's what I did in my test branch.

dtreeviz/models/shadow_decision_tree.py
parrt: Hmm...ok, let me think about this. the arg seems like a kludge/hack.

tlapusan: maybe having different methods for prediction and path ?

parrt: Why does get node samples call predict?

parrt: Is there no way to find samples w/o walking the tree very time? i see you caching them but maybe we compute this once up front for any tree.

tlapusan: predict() by default returned both prediction and prediction path nodes. To calculate the node samples, for lightgbm, we need to call predict() to get the nodes for each sample prediction. In this way we can calculate the samples for each node for lightgbm.

the challange is that the other libraries have different implementation to get the node samples:
- for sklearn, we can get the path directly for its API, dec_paths = self.tree_model.decision_path(self.x_data)  
- for spark, the leaf predictions is stored in tree metadata, so we dont need to re-calculate the leaf prediction from its node samples (like for lightgbm)

An alternative would be to have two separate functions, one for prediction and one for prediction path, but we will double the computational cost, which can be significant when we have a large number of samples (training)

I think the confusion is also caused by the fact the name of the method is predict() and in lightgbm case it will return only the prediction path.

tlapusan: @parrt would be a solution to handle this design issue in another PR ? (because it is related not only to lightgbm).
 
in the end, having two separate functions for prediction and paths, it will not increase the computational cost significantly, because we don't need both of values in costly functions. 
For example, the get_node_samples() is very costly because it handles all the training set examples, but it needs only the paths. The other functions which need both values are called for just one training example.

tlapusan: @parrt any thoughts on this ? :)

parrt: I agree let's merge and then please do create another PR and let's separate the function into two.

PR #115, opened 2020-12-07
make xgboost optional

oegedijk:
Hi there,

Made a few changes to make the xgboost import optional:

- wrap imports in try except blocks
- if `xgboost` import fails raise `ImportError`
- detect booster with `str(type(tree_model))` instead of `isinstance(tree_model, Booster)`
- added `extras_require` parameter to setup.py


parrt:
We'll need a general solution for all packages I'd say. Thanks for submitting.

tlapusan:
@oegedijk I've made a PR on your fork project with few changes

oegedijk:
did the equivalent to pyspark. Not sure whether should do the same thing to sklearn, as it also gets used in `trees.py` (not sure how essential it is there. Seems it mainly used for type annotation.)

parrt:
I think it's okay to assume sklearn.

tlapusan:
hi @oegedijk, I will take a look on your last commits and in the the next days I will come back to you. sorry for late responding :)

tlapusan:
I've looked over your changes and they worked for me also. The code is working if we dont have installed xgboost or pyspark. 

@parrt can you take a look also and to merge this PR ?

@oegedijk can you give us some pip examples of installing dtreeviz, without pyspark or xgboost. It will help also to have this pip examples specified in readme.md file. Thanks

dtreeviz/models/shadow_decision_tree.py
```diff
@@ -6,7 +6,13 @@
 import numpy as np
 import pandas as pd
 import sklearn
-import xgboost
+
```
parrt, line 9: seems like we can get rid of this dead code

dtreeviz/models/shadow_decision_tree.py
```diff
@@ -438,17 +444,26 @@ def _get_y_data(y_data):
 
     @staticmethod
     def get_shadow_tree(tree_model, x_data, y_data, feature_names, target_name, class_names=None, tree_index=None):
+        if hasattr(tree_model, 'get_booster'):
```
parrt, line 447: I'm not sure what this is doing?  Are we trying to do two things at once? merge XGBoost and Make them optional? Why wasn't this in there before is my question if we already support XGBoost

dtreeviz/models/shadow_decision_tree.py
```diff
@@ -438,17 +444,26 @@ def _get_y_data(y_data):
...
+            tree_model = tree_model.get_booster() # support XGBClassifier and XGBRegressor
         if isinstance(tree_model, ShadowDecTree):
             return tree_model
         elif isinstance(tree_model, (sklearn.tree.DecisionTreeRegressor, sklearn.tree.DecisionTreeClassifier)):
             from dtreeviz.models import sklearn_decision_trees
             return sklearn_decision_trees.ShadowSKDTree(tree_model, x_data, y_data, feature_names,
                                                         target_name, class_names)
-        elif isinstance(tree_model, xgboost.core.Booster):
+        elif str(type(tree_model)).endswith("xgboost.core.Booster'>"):
```
parrt, line 455: I'd have to check this myself but this seems like a complicated expression that we could simplify

dtreeviz/models/spark_decision_tree.py
```diff
@@ -6,8 +6,13 @@
 import pyspark
 
 from dtreeviz.models.shadow_decision_tree import ShadowDecTree
-from pyspark.ml.classification import DecisionTreeClassificationModel
-from pyspark.ml.regression import DecisionTreeRegressionModel
+try:
```
parrt, line 9: This file is specifically about spark so why can't we just do the import? why do we need try/catch.

parrt (commented):
sorry.  I clicked the wrong button and I can't seem to get rid of this.

dtreeviz/models/xgb_decision_tree.py
```diff
@@ -4,12 +4,18 @@
 from typing import List, Mapping
 
 import numpy as np
-import xgboost as xgb
-from xgboost.core import Booster
 
 from dtreeviz.models.shadow_decision_tree import VisualisationNotYetSupportedError
 from dtreeviz.models.shadow_decision_tree import ShadowDecTree
 
+try:
```
parrt, line 11: if this is specific to xgb  we should be able to just import it right? the import itself is guarded in other files.

parrt:
howdy.  I added a number of comments and, by mistake, created a single comment review.  Thanks very much for your efforts.

dtreeviz/models/shadow_decision_tree.py
oegedijk: This basically allows users to pass the sklearn compatible wrappers `xgboost.sklearn.XGBClassifier` and `xgboost.sklearn.XGBRegressor`. Both have a `get_booster()` method to get the underlying xgboost model. 

Otherwise users would just get weird error messages and have to figure out they have to do `.get_booster()` themselves.

dtreeviz/models/shadow_decision_tree.py
oegedijk: Could do something like [safe_isinstance()](https://github.com/slundberg/shap/blob/8a45ff574fb0cad4c2d3cee8e5a825c7ae5f82d8/shap/utils/_general.py#L172) like in the shap library for a slightly more elegant way of checking instance type without actually loading a module.

dtreeviz/models/spark_decision_tree.py
oegedijk: I think this is correct.

dtreeviz/models/xgb_decision_tree.py
oegedijk: idem

dtreeviz/models/shadow_decision_tree.py
tlapusan: indeed, it was a little confuse to me also... but it's the code to support https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBClassifier, which is a sklearn API for xgboost.  
@oegedijk can you confirm this ?

If this line of code was for you also confusing, we should refactor it somehow

oegedijk:
with regards to installs, you still install it as before: `pip install dtreeviz`, only now by default xgboost and pyspark do not get installed. So you can either do that yourself manually, or use the bracket expression `pip install dtreeviz[xgboost]` or `pip install dtreeviz[pyspark]`. I expect that most people that want to visualize an xgboost or pyspark tree already have the package installed however :)

dtreeviz/models/shadow_decision_tree.py
oegedijk: a more explicit way of doing it would be `if str(type(model)).endswith("XGBClassifier'>"): model=model.get_booster()`

dtreeviz/models/shadow_decision_tree.py
tlapusan: since str(type(tree_model)) is used in many places, we could create from it a variable named tree_model_type, for example, and use it like `elif 'xgboost.core.Booster' in tree_model_type:`

dtreeviz/models/spark_decision_tree.py
tlapusan: this was actually my suggestion. It serves more like a  warning message for the user... if he installed dtreeviz without pyspark and latter needed this code, to have this message... 

if I think twice, it doesn't add too much value.... we can refactor it if the code became too verbose using it.

dtreeviz/models/shadow_decision_tree.py
parrt: yes maybe a more explicit version is better. I think what really is lacking is a simple comment indicating what that code does ;)

parrt: Also instead of `str(type(...))` I wonder if it's more readable or explicit anyway to use `model.__class__.__name__`??

dtreeviz/models/spark_decision_tree.py
parrt: ah. I see.  That's a good thought, but I think if Python tells them there is no pyspark, they will get the message.  Might be better to simplify the code.

dtreeviz/models/shadow_decision_tree.py
parrt: Probably okay just to reference the `__name__` via its type.

parrt:
ok, guys.  I have resolved the conversations just so that I know I've looked at it. Can you make the requested changes? thanks!

oegedijk:
Removed the import error warning, and added a more detailed comment for the `.get_booster()` line.

I kept the `str(type(model)).endswith(...)` as it is more expressive than `model.__class__.__name__` which would just return `Booster` for the xgboost model instead of `xgboost.core.Booster`.

oegedijk:
also added the pyspark models to the incompatible models warning

oegedijk:
sorry for delay,

oegedijk:
Sorry, hit the wrong button and accidentally closed it :)

parrt:
@tlapusan  this one seems okay now. Thanks, @oegedijk !

tlapusan:
thanks @oegedijk !
I hope that now you can install your library, explainerdashboard, on heroku ;)

Issue #120, opened 2021-01-09
Using dtreeviz with sklearn pipelines


windisch:
When the following `model`

```
from sklearn.impute import KNNImputer
from sklearn.pipeline import make_pipeline
from sklearn.tree import DecisionTreeRegressor

model = make_pipeline(KNNImputer(), DecisionTreeRegressor())
```
is plugged into the function `dtreeviz.trees.dtreeviz`, this exception is raised:

```
ValueError: Tree model must be in (DecisionTreeRegressor, DecisionTreeClassifier, xgboost.core.Booster, but was Pipeline
```

Are there any plans to make sklearn's [pipelines](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html) work for `dtreeviz`?

parrt:
Hi.  well a pipeline could have 30 different elements in it... it's up to you to pull out the specific model you'd like to visualize.

windisch:
Thanks for your quick response! 

In my toy model above, the decision tree is at the end and cannot be pulled out of the pipeline as it needs all prior steps (in the toy example, only the imputing, but there may be more) to be executed. An example which may highlights my point is the following:

```
from sklearn.impute import KNNImputer
from sklearn.pipeline import make_pipeline
from sklearn.tree import DecisionTreeRegressor
from sklearn.preprocessing import PolynomialFeatures

model = make_pipeline(KNNImputer(), PolynomialFeatures(degree=2), DecisionTreeRegressor())
```

Here, pulling out the `DecisionTreeRegressor` would change the model, as  the decision tree in `model` operates on a totally different features space than a decision tree fitted on data without the `PolynomialFeatures` preprocessor before. 

I'm happy to help implementing the pipeline support. I was just wondering whether there exist already plans to implement this.

parrt:
```
t = DecisionTreeRegressor()
model = make_pipeline(KNNImputer(), PolynomialFeatures(degree=2), t)
```

windisch:
If I understand you correctly, it should work with the existing implementation. I don't understand how.

Using like that
```
from sklearn.impute import KNNImputer
from sklearn.pipeline import make_pipeline
from sklearn.tree import DecisionTreeRegressor
from sklearn.preprocessing import PolynomialFeatures
from sklearn.datasets import load_boston

from dtreeviz.trees import dtreeviz

t = DecisionTreeRegressor()
[...]

PR #124, opened 2021-01-31
Add helper to make dtreeviz work with pipelines

windisch:
This change adds a helper function as suggested by @tlapusan to extract the necessary input arguments for `dtreeviz` when the decision tree model used is part of a pipeline (see #120).

As wished, I added a small Jupiter notebook (basically a copy of the existing ones) showing how to use the helper. It might also be valid to have another section in the existing Jupiter notebook dealing with pipelines, but as this notebook is already kind of comprehensive, I added a new one.


tlapusan:
hi @windisch,
I've looked over your implementation and ran the code from the new notebook. It looks great !

In test module, do we need two initializations for features_expected ? their content seem to be the same.
The path for tests has been changed in upstream master. Can you merge it locally and to see that everything is ok ?

@parrt if you are ok with this PR, we can merge it.

thanks again @windisch for your efforts !

PR #127, opened 2021-02-06
Add random forest 2D feature space visualization

parrt:
I added two interesting methods that show the 2D feature space for classifiers associated with one or more trees. One shows the regions fused together, ctreeviz_bivar_fusion(), and the other does a nice tiling to show probabilities or predicted values for the grid, crfviz_bivar().


parrt:
Heh, @tlapusan  can you take a look at this and the notebook?  This is my first draft at the API but I think it's okay. Would be very happy to hear about recommendations for the interface or its functionality. It has a number of limitations at this point, such as being limited to sklearn or any other model that can answer method predict_proba().   Also assumes continuous class target values 0..k-1.

tlapusan:
cool, I will take a look soon.

dtreeviz/forests.py
```diff
@@ -0,0 +1,246 @@
+import numpy as np
+import pandas as pd
+
+import matplotlib.patches as patches
+import matplotlib.pyplot as plt
+from matplotlib.collections import PatchCollection
+from colour import Color
+from PIL import ImageColor
```
tlapusan, line 8: how did you manage to install PIL ? I have this error "ERROR: No matching distribution found for PIL"... 
maybe we should add it into setup.py (same for rfpimp from notebook)

dtreeviz/forests.py
```diff
@@ -0,0 +1,246 @@
...
+                          alpha=.1,
+                          ax=None):
+    if ax is None:
+        fig, ax = plt.subplots(1, 1, figsize=(4, 3.5))
+    ax.spines['top'].set_visible(False)  # turns off the top "spine" completely
+    ax.spines['right'].set_visible(False)
+    ax.spines['left'].set_linewidth(.5)
+    ax.spines['bottom'].set_linewidth(.5)
+
+    class_names = ShadowDecTree.normalize_class_names(class_names, nclasses=len(np.unique(y)))
```
tlapusan, line 29: since normalize_class_names() is used both for trees and random forests, would be an idea to extract this method in a util module ?

dtreeviz/forests.py
parrt: oh right. yeah, that is weird stuff. I think it's actually `Pillow` not `PIL` to install. Can you try that package?

parrt: actually I think they are different. Hmm...

parrt: Yep, different but we still want Pillow:

```
beast:RF-classifier-regions:~/github/dtreeviz $ pip uninstall PIL
WARNING: Skipping PIL as it is not installed.
beast:RF-classifier-regions:~/github/dtreeviz $ pip install -U Pillow
Requirement already satisfied: Pillow in /Users/parrt/opt/anaconda3/lib/python3.8/site-packages (8.1.0)
```

parrt: I updated setup.py and pushed a new commit.

dtreeviz/forests.py
parrt: that's a good idea. I'll move it.

tlapusan: ShadowDecTree.normalize_class_names is not defined anymore :)

dtreeviz/models/shadow_decision_tree.py
```diff
@@ -390,14 +392,7 @@ def get_leaf_sample_counts_by_class(self):
 
     def _normalize_class_names(self, class_names):
```
tlapusan, line 393: Could we remove this private function and use the one define in utils ?

dtreeviz/utils.py
```diff
@@ -125,6 +125,18 @@ def myround(v,ndigits=2):
     return format(v, '.' + str(ndigits) + 'f')
 
 
+def _normalize_class_names(class_names, nclasses):
```
tlapusan, line 128: _function_name(), single underscore is a hint for private/internal usage. Since this method is used externally, I guess we should remove the _

setup.py
```diff
@@ -4,11 +4,11 @@
 
 setup(
     name='dtreeviz',
-    version='1.1.4',
+    version='1.2',
     url='https://github.com/parrt/dtreeviz',
     license='MIT',
     packages=find_packages(),
-    install_requires=['graphviz>=0.9','pandas','numpy','scikit-learn',
+    install_requires=['graphviz>=0.9','Pillow','pandas','numpy','scikit-learn',
```
tlapusan, line 11: I will put this comment here because I cannot on rf-partitioning.ipynb. Nice visualisations btw ;)

pltvid isn't installed and after installing it, I have this error :
Screen Shot 2021-02-16 at 10 44 58 AM

dtreeviz/forests.py
tlapusan: @parrt I can solve this issues if you dont have time ?

setup.py
tlapusan: I managed to solve this by "brew install poppler" (it took a while to install it).
Do you remember how you managed to make it work for yourself? because by installing the poppler using brew can be uncomfortable. I saw on stackoverflow by installing it using pip, but didn't work...

dtreeviz/forests.py
tlapusan: true, we have created and put the normalize_class_names in utils module, but there where still few old references like ShadowDecTree.normalize_class_names. 

Created a PR to fix it.

dtreeviz/models/shadow_decision_tree.py
parrt: Oh right.  if I have moved to it then yes, let's remove :)

dtreeviz/utils.py
parrt: Ah.  I was using it to indicate for use inside of our library, although not necessarily limited to the module. I just wanted people to avoid using it outside dtreeviz.

parrt: I don't really have a preference; whatever you think is best

setup.py
parrt: actually I think that it is PIL that we need to install not Pillow, which is something else.

parrt:
I'm going to leave this alone for a while until I can finish some other projects; let's focus on getting your PR merged and released... I think I'm still behind on your LightGBM PR. going to look for that one now

parrt:
I had a commit that had the rename for _normalize ...  just pushed that.

Also I realize that it's actually Pillow not PIL. in fact the documentation for pillow says we need to remove PIL before installing pillow.

parrt:
@tlapusan I think we can close this one right? I think we have the two dimensional future space visualization in.

tlapusan:
to be honest, I don't remember too much but I see that the code was merged into master.  So we should be ok to close it.

PR #130, opened 2021-02-24
Import ABC from collections.abc for Python 3.10 compatibility.

### LINKED ISSUE #128: shadow_decision_tree.py: importing from collections raises a DeprecationWarning in Python 3.9
Hi there!

Running the tests gave the following warning:
```
  /usr/lib/python3/dist-packages/dtreeviz/models/shadow_decision_tree.py:2: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working
    from collections import Sequence
```

The solution is to replace this line with:
```python
    from collections.abc import Sequence
```

tirkarthi:
Fixes #128

### DIFF
--- a/dtreeviz/models/shadow_decision_tree.py
+++ b/dtreeviz/models/shadow_decision_tree.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from collections import Sequence
+from collections.abc import Sequence
 from numbers import Number
 from typing import List, Tuple, Mapping
 


tlapusan:
thanks @tirkarthi for this PR. 
I've checked it and also the unit tests are working.

We are using also defaultdict from collections, but it doesn't need this update.

@parrt from my side, it's ok to merge.

parrt:
Are we sure collections.abc works in 3.6? maybe we change to 3.7?

tlapusan:
On my laptop I have python 3.7... Could be an idea to change to python requirement >= 3.7

parrt:
I found a version of 3.6 and this worked `from collections.abc import Sequence`!

tirkarthi:
It should work from 3.4 and above.

PR #135, opened 2021-03-21
Warning for tree index

### LINKED ISSUE #134: Warning for tree index
Was trying for LGBM regression tree. Huge thanks for this new feature. Forgot to add tree_index parameter, and couldn't see a warning that emerges me to specify tree index. 

This gave me: 

viz = dtreeviz(model,
               train[model_features] ,
               train[target] ,
               target_name='Cons',
               feature_names=np.array(model_features))
 
This:
---> 58         tree_dump = self.booster.dump_model()["tree_info"][self.tree_index]
     59         _walk_tree(tree_dump["tree_structure"], node_index)
     60 

TypeError: list indices must be integers or slices, not NoneType_

Since tree_index not defined on function, this error happens. It should be a better a warning on specifying tree index.

tlapusan:
close #134

### DIFF
--- a/dtreeviz/utils.py
+++ b/dtreeviz/utils.py
@@ -174,6 +174,13 @@ def extract_params_from_pipeline(pipeline, x_data, feature_names):
     return tree_model, x_data, feature_names
 
 
+def check_tree_index(tree_index, nr_of_trees):
+    if tree_index is None:
+        raise Exception("You need to setup tree_index parameter")
+    if tree_index >= nr_of_trees:
+        raise Exception(f"tree_index parameter should have values between [{0}, {nr_of_trees - 1}]")
+
+
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


dtreeviz/utils.py
```diff
@@ -174,6 +174,13 @@ def extract_params_from_pipeline(pipeline, x_data, feature_names):
     return tree_model, x_data, feature_names
 
 
+def check_tree_index(tree_index, nr_of_trees):
+    if tree_index is None:
+        raise Exception("You need to setup tree_index parameter")
```
parrt, line 179: Is that the right exception? Maybe `ValueError`?

parrt, line 179: "setup"->"pass in a"

parrt (approved):

PR #139, opened 2021-03-29
Split predict method in predict and predict_path

tlapusan:
it seems that  the old usage of predict() method was only for prediction path...

Changed the unit tests, ran them and compared visually all the notebooks


parrt:
thanks!

PR #140, opened 2021-03-30
Class boundary viz

parrt:
Heh @tlapusan !  thinking 2D classifier probability visualizer should be separate PR.  This is getting closer.  Take a peek?  See [notebook](https://github.com/parrt/dtreeviz/blob/class-boundary-viz/notebooks/classifier-decision-boundaries.ipynb)


tlapusan:
hi @parrt, the new visualizations are looking really cool ! 

There is not so much clarity when the markers are overloaded, maybe a little transparency it will help.

Screenshot 2021-04-01 at 13 48 12

tlapusan:
after running the notebook locally, I've encounter some issues :
1. I've installed pltvid using pip
Screenshot 2021-04-01 at 14 02 41

2. 
Screenshot 2021-04-01 at 14 05 21

3. 
Screenshot 2021-04-01 at 14 06 46

4. after fixing the syntax error -> "name 'cyl4' is not defined"
Screenshot 2021-04-01 at 14 07 10

parrt:
oh right.  I had not played with the animations yet.   I have added tile_alpha as a parameter and it definitely helps.  I said it 2.8 by default now.  Also make sure you pull the latest because I moved that class name normalization to utils in the mainline. Working on the 1D version at moment. :)   It will look very similar to the ctreeviz_univar

parrt:
@tlapusan  just added all kinds of crazy stuff.  You can check out the new notebooks that start with `classifier-`. If it looks okay I will push on Monday or Tuesday as 1.3.   There's a unified `clfviz()` that will work with wanted two-dimensional data. Happy to have any comments on the code as well.

tlapusan:
I've looked over classifier-decision-boundaries notebook.

Do you consider having the legend as a default values in the show parameter ? I think it will help to make association between colors and target classes.

I like the idea of showing the misclassified examples.

For univar, are the misclassified examples showed correctly ? shouldn't yellow circles from the first green rectangle to be misclassified ?
Screenshot 2021-04-04 at 18 49 54

I have the same issues for classifier-boundary-animations notebook, "PDFInfoNotInstalledError: Unable to get page count. Is poppler installed and in PATH?". I do remember that I solved this issue on my previous laptop by installing some library using brew...still working on this.

parrt:
Hm...yeah, some more should be red. re legend: hmm...yeah I guess by default. sure!  oh right poppler...i wonder why it's needed now...

parrt:
Oh, i'm using randomforest which overfits like crazy...so it can isolate that weird one no problem.

parrt:
I did:

```
beast:class-boundary-viz:~/github/dtreeviz $ brew uninstall poppler
Uninstalling /usr/local/Cellar/poppler/21.02.0... (474 files, 27.0MB)
```

And reran notebooks with no problem.

parrt:
ok, poppler needed for animation using my pltvid lib.

parrt:
Ok, updated notebook to say needs poplar. Updated pltvid to work on windows. I think it's ready!

tlapusan:
hi @parrt, after some installs/uninstalls, I managed to make that error to disappear. 
Should camera.save(....) generate some images in the notebook ? because I have no output, like bellow.
Screenshot 2021-04-05 at 20 47 17

If the notebooks work as expected for you, I'm ok to make the new release, bacause the problem could be on my side (to many installs/uninstalls maybe)

parrt:
Hi. it just saves. doesn't show in notebook.

PR #147, opened 2021-05-27
Display just a range of tree levels

### LINKED ISSUE #143: memory error for big big trees
Hi Terence,

first of all, thank you for this amazing library (and for the great blog articles you put out there). 
I'm writing here because I experience an issue with a tree that has a big depth. 
Would it be possible to somehow tell dtreeviz that we want to plot only the first part of such a tree (say, the first four splits) ? That would help avoid memory errors when the tree is too big and the structure gets out of hand.

Thank you for your time

tlapusan:
close #143


parrt:
hiya. why not use just a max depth rather than tuple?

tlapusan:
the first version was with only the max depth, but after that I thought that would be useful to visualise ranges of levels, for example the last n levels from the tree (containing the leaves)

dtreeviz/trees.py
```diff
@@ -563,6 +564,7 @@ def dtreeviz(tree_model,
                             display only those features
                            used to guide X vector down tree. Helps when len(X) is large.
                            Default is 25.
+    :param depth_range_to_display: range of depth levels to be displayed.
```
parrt, line 567: might add that ranges are inclusive

dtreeviz/trees.py
```diff
@@ -859,6 +867,9 @@ def get_leaves():
     edges = []
     # non leaf edges with > and <=
     for node in get_internal_nodes():
+        if depth_range_to_display is not None:
+            if node.level not in range(depth_range_to_display[0], depth_range_to_display[1]):
```
parrt, line 871: do we need `+1` ?

tlapusan: this loop is for displaying the edges, so we dont need to display the edges for the last level

parrt: ah!

parrt:
Thanks!

PR #153, opened 2021-08-05
cjk support and examples

alitrack:
use utils.default_font to easily support CJK


tlapusan:
hi @alitrack,

I've tried your example with default_font = plt.rcParams['font.sans-serif'][0] OR default_font="Arial" and the dtreeviz plots loo the same.

Could you please describe what's the value of having default_font = plt.rcParams['font.sans-serif'][0] ? thanks.

alitrack:
I want to render Chinese or CJK characters, so I need font to supports it.

feel free to have a try https://github.com/alitrack/dtreeviz/blob/master/notebooks/cjk.ipynb

tlapusan:
I've tried the default_font="Arial" and the dtreeviz from your notebook is working.

Screenshot 2021-08-05 at 20 01 27

Do you see any issues in the above visualisation ? Or could you make a screenshot with what is wrong displayed in case of Arial font ? Thanks

alitrack:
Arial does not support CJK characters

image


and many warning like these,
```
/Users/steven/anaconda3/envs/py39/lib/python3.9/site-packages/matplotlib/textpath.py:65: RuntimeWarning: Glyph 23665 missing from current font.
  font.set_text(s, 0.0, flags=LOAD_NO_HINTING)
/Users/steven/anaconda3/envs/py39/lib/python3.9/site-packages/matplotlib/textpath.py:65: RuntimeWarning: Glyph 40482 missing from current font.
  font.set_text(s, 0.0, flags=LOAD_NO_HINTING)
```

please run the following code,
```
from matplotlib import pyplot as plt
plt.rcdefaults()
print(plt.rcParams['font.sans-serif'])
```
my output is ,
```
['DejaVu Sans', 'Bitstream Vera Sans', 'Computer Modern Sans Serif', 'Lucida Grande', 'Verdana', 'Geneva', 'Lucid', 'Arial', 'Helvetica', 'Avant Garde', 'sans-serif']
```
image

the default font is DejaVu Sans, it does not support CJK characters either.

tlapusan:
indeed, I've reproduce the issue, don't know why it worked yesterday.... maybe the auto reload didn't work.

the output of plt.rcParams['font.sans-serif'] looks the same like yours. But I observed that you put the "Arial Unicode MS" in the head list into the notebook :
`#first, you have a font supports cjk, for example, Arial Unicode MS(Arial Unicode.ttf)
default_font = 'Arial Unicode MS'
from matplotlib import pyplot as plt
plt.rcParams['font.sans-serif']=[default_font]+plt.rcParams['font.sans-serif']
plt.rcParams['axes.unicode_minus']=False # in case minus sign is shown as box`

Would make more sense to put these lines of code in the utils.py ? or to simple define default_font="Arial Unicode MS" ? Thanks

alitrack:
in fact, `Arial Unicode MS` is just an example, and I am not sure if the system  has this font。
and I can use any font I want.
for example, i use a CJK font without install it

```python
from matplotlib import pyplot as plt,font_manager as fm
from pathlib import Path
import os
#Restore the `.rcParams` from Matplotlib's internal default style.
plt.rcdefaults()

path = Path(os.getcwd())
fname=os.path.join(path.parent.absolute(),'data','NotoSerifCJK-Regular.ttc')
fontProperties=fm.FontProperties(fname=fname,size=14)
default_font=fontProperties.get_name()# "Arial Unicode MS"
if default_font not in [f.name for f in fm.fontManager.ttflist]:
    print(f"{default_font} does not exist, let's add it to fontManager" )

if fname not in [f.fname for f in fm.fontManager.ttflist]:
    fm.fontManager.addfont(fname) # need absolute path

plt.rcParams['font.sans-serif']=[default_font]+plt.rcParams['font.sans-serif']
plt.rcParams['axes.unicode_minus']=False # in case minus sign is shown as box
```
I have put example [here](https://github.com/alitrack/ml)  , includes
Chinese(CJK) support without install font and  how to use dtreeviz with Streamlit.

parrt:
Hi @alitrack  Thanks for this PR.  Can you help me understand why we need to alter the default font mechanism when you can simply pass in the font you want as an argument to the functions?

alitrack:
utils.default_font is not necessary,  

```python 
    for text in leg.get_texts():
        text.set_color(colors['text'])
        text.set_fontsize(fontsize)
```
does not set fontname, will get it from plt.rcParams['font.sans-serif'], not utils.default_font

parrt:
Sorry. I don't follow.  We have a function fontname parameter. Why can't you pass that to get Chinese char?

alitrack:
in you code, there are many place use plt.rcParams['font.sans-serif'] as default font, because you did not use your default font, 
another way , forget my pull request, and try my modified version  [trees.py](https://gist.github.com/alitrack/2ad6c2c84b78bcf11402af7ea16f68ee#file-trees-py)  and you can also get correct plot(but maybe does not work for other code).

```python
from sklearn.datasets import *
from sklearn import tree
from dtreeviz.trees import *

classifier = tree.DecisionTreeClassifier(max_depth=3)  # limit depth of tree
iris = load_iris()
classifier.fit(iris.data, iris.target)
# the following two line codes are just for Chinese test.
feature_names = ['萼片sepal length (cm)','萼片sepal width (cm)',
                 '花瓣petal length (cm)','花瓣petal width (cm)']
class_names = ["山鸢尾 (Iris-setosa)", "变色鸢尾(Iris-versicolor)", "维吉尼亚鸢尾(Iris-virginica)"]
viz = dtreeviz(classifier,
               iris.data,
               iris.target,
               target_name='variety',
               fontname='Arial Unicode MS',
               feature_names=feature_names,
               class_names=  class_names # need class_names for classifier
              )

viz
```

here are some issues,

`draw_piechart`  and `draw_legend` doe not have fontname parameter, and so on  does not use the font you set, but plt.rcParams['font.sans-serif']

`add_classifier_legend` has fontname parameter, but you does not give it your default font name.

parrt:
Thank you very much for the explanation. @tlapusan and I can investigate and find simplest solution.

tlapusan:
Indeed, the fontname parameter wasn't sent to all the methods. I also think that fixing this and passing the desired value for fontname as a parameter is the right approach.

I've made few changes and now the plot is working correctly.
Screenshot 2021-08-09 at 10 00 20

@alitrack if you have more examples of methods which are not using the right fontname, it will be helpul if you can post them here :) thanks.

tlapusan:
@alitrack you can follow this PR #154 if you are still interested ;)

tlapusan:
@parrt, I think we can close this PR. We've just merge yesterdat the solution into master from another PR.

PR #154, opened 2021-08-12
Use fontname sent as parameter

tlapusan:
Tried to identify all methods where fontname parameter could be used.
@parrt if you can take a look also, would be helpful. The examples from #153 works if you specify the fontname = "Arial Unicode MS".


tlapusan:
@parrt FMPV this PR can be merged. do you still want to check it more ?

parrt:
Awesome. thanks!

PR #161, opened 2021-10-28
feature: ranges of plot can be set manually

DJCordhose:
when exploring how your model extrapolates it can be useful to extend the plot of decision boundaries way beyond min and max of given data. this is now possible by manually specifying ranges


parrt:
Thanks, Ollie!

PR #180, opened 2022-03-08
Fix scale_SVG

borjagarciag:
This function replaces the scale of the SVG with a given input. SVG file uses float numbers, but this function searched for integer numbers, which did not exist.


parrt:
Wonder why it worked before. Hmm...anyway, examples work. thanks!

Issue #26, opened 2018-12-03
dtreeviz fails with IndexError for n_classes >= 11


CermakM:
Apparently `dtreeviz` in `dtreeviz.trees.py` fails on the following block of code:

```python
> /home/macermak/.local/lib/python3.6/site-packages/dtreeviz/trees.py(320)dtreeviz()
    318 
    319     n_classes = shadow_tree.nclasses()
--> 320     color_values = color_blind_friendly_colors[n_classes]
    321 
    322     # Fix the mapping from target value to color for entire tree
```

resulting in

```python
IndexError: list index out of range
```

My model has __`15` classes__ and I believe the length of `color_blind_friendly_colors` (which is __`11`__) is causing the issue.

parrt:
ah. yeah, sorry. I should put a check in there but only 10 classes can be handled right now.  we had to handpick the colors.  sorry about that.

CermakM:
Wow, quick response!

So there is nothing I can do at the moment?

parrt:
Nope :( sorry.

CermakM:
Okay, let's hope for a quick solution in the bright future. Keep up the good work and thanks for the quick response.

parrt:
thanks :)  I'd be happy to accept a pull request that increased the number of color palettes ;)

Ashafix:
It's a bit more tricky than just having more colors.

`get_num_bins` gets called with `n_classes` as well and it references `NUM_BINS` which also only has 10 values.
What would be a good default value for `NUM_BINS` for `n_classes` > 10?

Any good ideas about having more colors? 
Just repeating the colors and giving a warning, or making up new colors?

parrt:
Yeah, That's the problem. I started running out of ideas for colors and default values above 10.  May be a sequential color map is required.

EricCacciavillani:
Currently running into the same issue for an idea I am testing. 
But totally understandable. Maybe a semi-simple and quick workaround is just to declare a simple target to color hex dict. 

`{0: #fcba03,
   1: #fc03a9}`

chenhajaj:
Any solution? I even considered not using colors, but when I add None to the color_blind_friendly_colors array (colors.py), I get the following error:
 color_map = {v: color_values[i] for i, v in enumerate(class_values)}

TypeError: 'NoneType' object is not subscriptable

parrt:
In principle, just add another list of colors so this works:

```
color_values = color_blind_friendly_colors[n_classes]
```

[...]

PR #185, opened 2022-06-08
added support for up to 40 colors

jlubbersgeo:
by adding a helper function in ```colors.py```, we can now create as many discrete colors from any valid matplotlib colormap as a classifier model has classes. The limit on this is 40 based on the length of ```NUM_BINS``` in ```trees.py``` which should take care of most classification problems. 

When calling the ```dtreeviz``` function the argument ```cmap``` can now be specified, otherwise defaults to the ```RdYlBu``` from matplotlib:

```python
viz = dtreeviz(
    rf.estimators_[tree_number],
    X,
    y,
    target_name="target_class",
    feature_names=features_list,
    class_names=class_names,
    cmap = 'viridis'

)
viz.view()
```

### DIFF
--- a/dtreeviz/colors.py
+++ b/dtreeviz/colors.py
@@ -1,3 +1,6 @@
+import matplotlib
+import numpy as np
+
 YELLOW = '#fefecd'
 GREEN = '#cfe2d4'
 DARKBLUE = '#313695'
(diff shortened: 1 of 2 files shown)


dtreeviz/trees.py
```diff
@@ -766,6 +778,10 @@ def get_leaves():
         highlight_path = [n.id for n in path]
 
     n_classes = shadow_tree.nclasses()
+
+    # call the get_hex_colors function for up to 40 classes
+    colors['classes'] = get_hex_colors(n_classes,cmap)
```
parrt, line 783: hi! I can't tell if this defaults to color blind reasonable colors. Does it keep same color scheme for <= 10 classes?

jlubbersgeo: This will do `n_classes` discrete colors  evenly spaced from the specified matplotlib colormap which doesn't necessarily mean that they will be colorblind friendly. Ultimately it would be on the user to choose a colormap that is colorblind friendly from something like the [perceptually uniform](https://matplotlib.org/3.5.0/tutorials/colors/colormaps.html#sequential) list of colors. Right now I have it set to default to the `RdYlBu` colormap because that seemed the closest matplotlib colormap to what the original settings were. 

Perhaps a conditional statement for if `n_classes` is less than or equal to 10 that would default to the originally specified colors would work well here?

parrt: This would change the basic color set then for every user I'd say.  Yeah, we need some thing that's backwards compatible. Maybe the conditional as you say would be enough.

tlapusan: I checked here how the viz is looking if number of classes > 10.
Screenshot 2022-06-13 at 10 33 22

I agree that we should keep backward compatibilities if n_classes <= 10. But if the n_classes > 10, we can implement add @jlubbersgeo solution. Better to render the dtreeviz than throwing some errors.  https://github.com/parrt/dtreeviz/issues/26 it's another issue complaining about this.

Regarding NUM_BINS limit of 40, we can fix it also. get_num_bins() is just a way to establish the number of bins for a histogram. We can see from the code that the latest values from the list are all of the same value "5". We can choose the same value "5" in case the n_classes > 40.

parrt: Sounds good. @jlubbersgeo can you add the `n_classes <= 10` check?

jlubbersgeo: @parrt yep, I'll take care of it early this week some time!

tlapusan: Sounds great!

jlubbersgeo: @parrt Sorry for the delay...some work/covid things came up, but I added a line to make everything <= 10 classes keep the original colorblind friendly colors.

tlapusan: Hi @jlubbersgeo, thanks and hope you are fine now :)

I checked your code and it works ok, but I observed a little issues (don't figure it out why)

Can you please check the following scenario ?
Try to print a tree with 2 classes, after that with classes > 10 and after that again with 2 classes.
In my case, when you print the tree second time with 2 classes, it doesn't reset the colors to the original, and takes the colors from the tree with classes > 10. Don't understand why :)

parrt: weird! Let me know if there's any progress and I can try myself and then merge...

jlubbersgeo: Looks like `COLORS["colors"]` was being permanently changed after any tree with > 10 classes was being created. Any tree after that even with < 10 classes was then drawing from that list of colors and not the original colorblind friendly color list. It should be fixed such that this doesn't happen. The following test produces the desired results:
```python
from dtreeviz.trees import *
from sklearn.datasets import make_classification
from sklearn.model_selection import train_test_split
from sklearn.ensemble import RandomForestClassifier

n_classes = [2, 10, 15, 5, 20]
for n in n_classes:
    X, y = make_classification(n_samples=1000, n_classes=n, n_informative=6)
    X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)
    clf = RandomForestClassifier(max_depth=3)
    clf.fit(X, y)
    tree_num = 20
    viz = dtreeviz(clf.estimators_[tree_num], X, y,cmap = 'viridis')
    viz.view()
```

parrt: @tlapusan do you have time to check? thanks!

parrt:
i pushed 1.3.7 just now...

dtreeviz/trees.py
jlubbersgeo: @tlapusan great news! Just building on an already great package. Looking forward to including some nice dtreeviz figures in my upcoming presentations/papers.

tlapusan: @jlubbersgeo sounds great ! Let us know if you are gonna publish them!

tlapusan: now it is pretty clear why the colors classes are overwritten. It's because the colors is a dictionary variable (like static in java), which has a general state and don't reset after each dtreeviz method call.

I'm working on a solution. Also I would like to move the logic of determining the colors values(that if n_classes > 10) inside the colors.py module. It's more appropiate to be there,  otherwise each visualisation method from the library would need to add that if logic.

dtreeviz/colors.py
```diff
@@ -24,7 +27,61 @@
     ['#FEFEBB', '#edf8b1', '#c7e9b4', '#7fcdbb', '#1d91c0', '#225ea8', '#fdae61', '#f46d43'],  # 8
     ['#FEFEBB', '#c7e9b4', '#41b6c4', '#74add1', BLUE, DARKBLUE, LIGHTORANGE, '#fdae61', '#f46d43'],  # 9
     ['#FEFEBB', '#c7e9b4', '#41b6c4', '#74add1', BLUE, DARKBLUE, LIGHTORANGE, '#fdae61', '#f46d43', '#d73027']  # 10
+
 ]
+def get_colorblind_friendly_colors():
```
parrt, line 32: @tlapusan did this get merged in? I don't see this in colors.py

tlapusan: @parrt it should be. Looking into the master code, we support the case when the number of classes is big. Implemented by get_hex_colors() method.

PR #186, opened 2022-07-13
Corrected flow of control of color classes

f-rosato:
Corrected a bug in the flow of control of dtreeviz. Now a potential user-supplied color dictionary would not have the content of the 'classes' key overwritten.

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -779,14 +779,11 @@ def dtreeviz(tree_model,
 
     n_classes = shadow_tree.nclasses()
 
-    # only generate custom colors if n_classes > 10
-    # otherwise keep the original colorblind friendly colors
-    if n_classes > 10:
+    # Only generate custom colors if the current class color lists,
+    # whether user-supplied or the default ones, are not enough.
+    if n_classes+1 > len(colors['classes']):
         # call the get_hex_colors function for up to 40 classes
         colors['classes'] = get_hex_colors(n_classes+1,cmap) 
-
-    else:
-        colors['classes'] = get_colorblind_friendly_colors()
     
     color_values = colors['classes'][n_classes]
 


parrt:
is it worth doing a release just for this fix or should we wait?

f-rosato:
I actually discovered and began using dtreeviz just yesterday (and happened to need this exact feature), so my opinion on the release schedule of the project would not be very informed...it seems to me it's small stuff that could wait for a little bit, but it's your call, I guess!

parrt:
ah. cool. @tlapusan I *think* this preserves original colors but I didn't check by running. haha

tlapusan:
@parrt @f-rosato 

seems to work, but it reintroduces the issue discussed with @jlubbersgeo here :https://github.com/parrt/dtreeviz/pull/185#discussion_r911684572

@f-rosato could you please provide us a user-supplied color dictionary, just to check with it :)

I guess is more important to fix the user supplied color dictionary than the issue mentioned in the above link. I will create an task issue with it based on how this conversation will follow.

f-rosato:
Sure, this is what I am using (4 shades of blue, as I have 4 classes max)

```
clr = ['#c1e7ff' ,'#86b0cc', '#4c7c9b', '#004c6d']
colors={'classes': [None, clr[:-3], clr[:-2], clr[:-1], clr]}
```

parrt:
Sorry about that @tlapusan. I got confused with the username and thought that he was correcting the recent colormap PR.

f-rosato:
Sorry if I caused confusion with my out-of-the-blue pull request 😅

tlapusan:
@f-rosato I got to reproduce your use-case. 

I think your fix is more important than what we fixed in https://github.com/parrt/dtreeviz/pull/185#discussion_r911684572

We can create an issue to don't forget it :)

PR #191, opened 2022-08-03
Support for tf df trees #181

### LINKED ISSUE #181: Support for tensorflow - decision forest
https://blog.tensorflow.org/2021/05/introducing-tensorflow-decision-forests.html

tlapusan:
@parrt it contains the visualisations for Random Forest algo. The tensorflow DF contains more algorithms for decision forests and we can add them also in the future.

close #181


testing/testlib/models/test_decision_tree_spark_classifier.py
```diff
@@ -21,10 +21,10 @@ def tree_model() -> (DecisionTreeClassificationModel):
 
 
 @pytest.fixture()
-def spark_dtree(tree_model, dataset_spark) -> ShadowSparkTree:
+def spark_dtree(tree_model, dataset_spark_tf) -> ShadowSparkTree:
```
parrt, line 25: Should this really be `tf`? Seems like it should be just spark or is somehow spark using tf underneath?

parrt:
hooray!!!  The `tfdf` prefix looks very much like the common term `tfidf`... Perhaps we change that to simply prefix of `tf` everywhere?

parrt:
There are a couple of changes such as in interpretation.py that might affect other targets. Have you run through and checked the other kinds of trees and run tests?

testing/testlib/models/test_decision_tree_spark_classifier.py
tlapusan: dataset_spark_tf it's a fixture for the dataset used in training for Spark and TF. I used the same dataset for testing for both spark and TF. Looking in the conftest.py I see we have different dataset fixtures for each library...don't remember exactly why.. maybe we needed a different format.

tlapusan:
> There are a couple of changes such as in interpretation.py that might affect other targets. Have you run through and checked the other kinds of trees and run tests?

yes, I spent some time for these changes and ran/tested all of them :)

tlapusan:
> hooray!!! The `tfdf` prefix looks very much like the common term `tfidf`... Perhaps we change that to simply prefix of `tf` everywhere?

agree, I will rename them soon.

parrt:
Thanks, @tlapusan !

PR #209, opened 2022-12-22
Add pip install for colab and change the colab link for each notebook

tlapusan:


parrt:
Looks great. By the way one of the issues I see is that I think SVG files do not show up on GitHub; should we be using retina/png in the notebooks so that they have images of hearing when you look on GitHub?

PR #210, opened 2022-12-22
Small cleanup of API

parrt:
don't assume a figsize; let it default to matplotlib size unless they set figsize as arg (or pass in an ax).


parrt:
looks simple enough...merging.

PR #215, opened 2022-12-28
Add option to plot classifier leaves as horizontal bars

mepland:
Added `leaf_plot_type` parameter. Default is `pie` which replicates prior plots, but now includes a `barh` option!

dtreeviz

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -1356,7 +1356,7 @@ def _draw_barh_chart(counts, size, colors, filename, label=None, fontname="Arial
             colors = [colors[i]]
 
     tweak = size * .01
-    fig, ax = plt.subplots(1, 1, figsize=(size, 0.3))
+    fig, ax = plt.subplots(1, 1, figsize=(size, 0.2))
 
     data_cum = 0
     for i in range(len(counts)):
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -223,6 +223,7 @@ class DTreeVizAPI:
              precision: int = 2,
              orientation: ('TD', 'LR') = "TD",
              instance_orientation: ("TD", "LR") = "LR",
+             leaf_plot_type: ('pie', 'barh') = 'pie',
              show_root_edge_labels: bool = True,
              show_node_labels: bool = False,
              show_just_path: bool = False,
@@ -255,6 +256,7 @@ class DTreeVizAPI:
                           after the decimal point. Default is 2.
         :param orientation:  Is the tree top down, "TD", or left to right, "LR"?
         :param instance_orientation: table orientation (TD, LR) for showing feature prediction's values.
+        :param leaf_plot_type: leaf plot type ('pie', 'barh')
         :param show_root_edge_labels: Include < and >= on the edges emanating from the root?
         :param show_node_labels: Add "Node id" to top of each node in graph for educational purposes
         :param show_just_path: If True, it shows only the sample(X) prediction path
@@ -554,7 +556,8 @@ class DTreeVizAPI:
                 _class_leaf_viz(node, colors=color_values,
                                 filename=f"{tmp}/leaf{node.id}_{os.getpid()}.svg",
                                 graph_colors=colors,
-                                fontname=fontname)
+                                fontname=fontname,
+                                leaf_plot_type=leaf_plot_type)
                 leaves.append(class_leaf_node(node))
             else:
                 # for now, always gen leaf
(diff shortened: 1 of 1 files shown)


dtreeviz/trees.py
```diff
@@ -1334,6 +1345,43 @@ def _draw_piechart(counts, size, colors, filename, label=None, fontname="Arial",
...
+    n_nonzero = np.count_nonzero(counts)
+
+    if n_nonzero != 0:
+        i = np.nonzero(counts)[0][0]
+        if n_nonzero == 1:
+            counts = [counts[i]]
+            colors = [colors[i]]
+
+    tweak = size * .01
+    fig, ax = plt.subplots(1, 1, figsize=(size, 0.3))
```
parrt, line 1359: This looks amazing. The bar charts look a little thick to me. I wonder if 0.2 would look better?

parrt:
How does the new height look? If you want I can just pull it in and tweak it on my side

mepland:
dtreeviz

New height looks good!

BTW my next commit should be signed with my ssh key.

parrt:
love it. thanks!

parrt:
@tlapusan should we make this bar chart the default leaf?  People seem convinced pie charts are bad haha.

parrt:
BTW, it occurs to me guys that we want the leaves to show relative area and not overlapping.  "stacked" `barh` is what we want I think but we'd want this non-overlapping regions.  E.g., if two classes are 10 and 10 samples each, then we don't want to hide one. We want equal visual regions. @tlapusan

mepland:
They don't overlap, due to the `left` param here:
```
        plt.barh(0, width, left=data_cum, color=colors[i], height=1, edgecolor=graph_colors['rect_edge'], linewidth=0.5)
```

Also originally I had the length of the bar plots scale with the total number of samples in the leaf, but we turned that off. Though I wouldn't mind bringing that back:
Screenshot 2022-11-12 at 13 29 12

parrt:
cool. yeah, i had the pie size scale a bit.  I think maybe scaling with n is good. @tlapusan?

Also we should remove the 0..57 x-axis labels; i think n=58 with the y class is best.

mepland:
Yeah we updated it to not have x axis ticks and just put the `n=...` below the bar. This was just the only old screenshot I had where the sized changed with n.

parrt:
ah, cool :)  yeah, i think it looks good:

Screenshot 2022-12-30 at 2 30 42 PM

parrt:
@mepland be advised that I did a rename of the argument: https://github.com/parrt/dtreeviz/commit/2004e8e5eed31306744337d66774384839f1467b

mepland:
> ah, cool :) yeah, i think it looks good:
> 
> Screenshot 2022-12-30 at 2 30 42 PM

@parrt did you change & commit the code to re-enable horizontal scaling with `n`? This looks great, I just don't see the change in code anywhere.

Also, I'm fine if this is the default behavior, but @tlapusan wanted them to all be the same size. Maybe we should make an option to control this behavior?

parrt:
Latest dev branch shows these two:

```
viz_model.view()
viz_model.view(leaftype='barh')
```
image
image

I think it's important to show where the mass of samples go in the leaves but I can imagine making this an option, thought it mirrors the current behavior with size now.

parrt:
I also think I'm going to make the bar charts the default, since most people have an immediate "but but but I was trying to the pie charts are bad" reaction.

parrt:
See: https://github.com/parrt/dtreeviz/commit/423060f3a70f3f83a2778bcc84412b2281db295c

tlapusan:
> @tlapusan should we make this bar chart the default leaf? People seem convinced pie charts are bad haha.

hmm, personally I like the pie charts :) highlights better the leaves and make the viz a little more interesting.
@parrt curious why the people think that the pies are a bad option.

If you heard this feedback many times, I'm ok to choose the barh as the default.

tlapusan:
> > ah, cool :) yeah, i think it looks good:
> > Screenshot 2022-12-30 at 2 30 42 PM
> 
> @parrt did you change & commit the code to re-enable horizontal scaling with `n`? This looks great, I just don't see the change in code anywhere.
> 
> Also, I'm fine if this is the default behavior, but @tlapusan wanted them to all be the same size. Maybe we should make an option to control this behavior?

@mepland I'm also for making the leaf size based on its input samples. Don't remember when I mentioned that I would like them to be the same size. Sorry if I did :)

mepland:
> @mepland I'm also for making the leaf size based on its input samples. Don't remember when I mentioned that I would like them to be the same size. Sorry if I did :)
No problem, I may have miss-understood.

parrt:
No problem. We can always back this out if @tlapusan has a good reason to fix the size. :)

PR #216, opened 2022-12-28
Add matplotlib default colors as mpl_colors

mepland:
Defined a set of matplotlib default colors, `C0, C1, C2`, ..., for convenience. See `mpl_colors`, which mirrors `color_blind_friendly_colors`.

We can leave the default colors as `color_blind_friendly_colors`, but I think it is useful to set this up in advance for users, so they don't need to themselves.

dtreeviz

### DIFF
--- a/dtreeviz/colors.py
+++ b/dtreeviz/colors.py
@@ -31,6 +31,17 @@ color_blind_friendly_colors = [
 ]
 
 
+mpl_colors = [
+    None,  # 0 classes
+    None,  # 1 class
+]
+for n_classes in range(2, 11):
+    class_colors = []
+    for i in range(0, n_classes):
+        class_colors.append(f'C{i}')
+    mpl_colors.append(class_colors)
+
+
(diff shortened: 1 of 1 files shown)


parrt:
(I change the base for you to dev) :)

dtreeviz/colors.py
```diff
@@ -31,6 +31,17 @@
...
+mpl_colors = [
+    None,  # 0 classes
+    None,  # 1 class
+]
+for n_classes in range(2, 11):
+    class_colors = []
+    for i in range(0, n_classes):
+        class_colors.append(f'C{i}')
+    mpl_colors.append(class_colors)
+
```
parrt, line 43: would it be more (or less) clear if we did:

```
mpl_colors = [
    None,  # 0 classes
    None,  # 1 class
] + [[f'C{i}' for i in range(0, n_classes)]
              for n_classes in range(2, 11)]
```

mepland: Looks fine to me!

PR #218, opened 2022-12-28
Fix digraph hard coding

mepland:
- Used `colors["text"]` instead of hard coded `#444443` in digraph.
- Added missing fontcolor="{text_color}" to digraph edges. Now font color will match matplotlib text.
- Use  `fontname`  instead of hard coded Helvetica, in digraph.

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -303,7 +303,7 @@ class DTreeVizAPI:
                     </tr>
                     </table>"""
             else:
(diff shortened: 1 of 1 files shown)


parrt:
hi @mepland ! I recently created a dev branch; Could you set the base to parrt:dev for your PRs? Thanks!  Oh, and I also just turned on the requirement to sign your commits, given how major the contributions are... I should probably make this more formal.

PR #219, opened 2022-12-28
Allow different left and right arrow colors via 'larrow' and 'rarrow'

mepland:
Allow different left and right arrow colors via `'larrow'` and `'rarrow'`. Existing `'arrow'` color parameter still works for both.


dtreeviz/trees.py
```diff
@@ -606,7 +606,13 @@ def get_leaves():
             if node.is_categorical_split() and not self.shadow_tree.is_classifier():
                 lcolor, rcolor = colors["categorical_split_left"], colors["categorical_split_right"]
             else:
-                lcolor = rcolor = colors['arrow']
+                if 'larrow' in colors and 'rarrow' in colors:
```
parrt, line 609: Seems like we should add these as formal definitions in the colors array in colors.py after:

```
          'arrow': GREY,
```

The specific keys override the general `arrow`.  What do we do when `larrow` is specified but `rarrow` is not? Seems like in that case `rarrow` should default to `arrow`.

mepland: We can add them in `colors.py`.

If one of `larrow` or `rarrow` is specified but the other is not then you should get the:
```
raise ValueError('Can not find arrow, or larrow AND rarrow in colors dict!')
```

I can change it to fall back to `arrow` if only one of the two is specified though, if you want.

mepland: @parrt please let me know what you'd like to do here.

parrt: What you have is great...forgot to hit "send" :)

parrt: But please do add entries to colors.py with `larrow: None` or won't that work.  That indicates the possibilities to users I think.

mepland: @parrt Done!

parrt:
Great work! Thanks :)

PR #220, opened 2022-12-28
ctree_leaf_distributions() label_all_leaves parameter 

mepland:
Added `label_all_leaves` parameter which turns on (off) x axis ticks for each leaf, useful when there are few (many) leaves.

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -147,7 +147,7 @@ class DTreeVizAPI:
                                  fontsize: int = 10,
                                  fontname: str = "Arial",
                                  grid: bool = False,
-                                 label_all_leaves: bool = True,
+                                 show_leaf_labels: bool = True,
                                  figsize: tuple = None,
                                  ax=None):
         """Visualize the distribution of classes for each leaf.
@@ -177,7 +177,7 @@ class DTreeVizAPI:
             Plot labels font name
         :param grid: bool
             True if we want to display the grid lines on the visualization
-        :param label_all_leaves: bool
+        :param show_leaf_labels: bool
             True if we want to label all leaves IDs on the x axis
         :param figsize: optional
             (width, height) in inches for the entire plot
@@ -199,12 +199,12 @@ class DTreeVizAPI:
             ax.spines['right'].set_visible(False)
             ax.spines['left'].set_linewidth(.3)
             ax.spines['bottom'].set_linewidth(.3)
-            ax.tick_params(axis='x', which='major', labelcolor=colors['tick_label'], top=False, bottom=label_all_leaves)
+            ax.tick_params(axis='x', which='major', labelcolor=colors['tick_label'], top=False, bottom=show_leaf_labels)
             ax.tick_params(axis='y', which='both', labelcolor=colors['tick_label'], left=True, right=False)
             ax.tick_params(axis='x', which='minor', labelcolor=colors['tick_label'], top=False, bottom=False)
 
             xlabel='leaf ids'
-            if label_all_leaves:
+            if show_leaf_labels:
                 ax.set_xticks(range(0, len(index)))
                 ax.set_xticklabels(index)
             else:
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -148,7 +148,8 @@ class DTreeVizAPI:
                                  fontname: str = "Arial",
                                  grid: bool = False,
                                  figsize: tuple = None,
-                                 ax=None):
+                                 ax=None,
+                                 label_all_leaves: bool = True):
         """Visualize the distribution of classes for each leaf.
 
         It's a good way to see how classes are distributed in leaves. For example, you can observe that in some
@@ -176,8 +177,12 @@ class DTreeVizAPI:
             Plot labels font name
         :param grid: bool
             True if we want to display the grid lines on the visualization
-        :param figsize: optional (width, height) in inches for the entire plot
-            :param ax: optional matplotlib "axes" to draw into
+        :param figsize: optional
+            (width, height) in inches for the entire plot
+        :param ax: optional
+            matplotlib "axes" to draw into
+        :param label_all_leaves: bool
+            True if we want to label all leaves IDs on the x axis
         """
         index, leaf_samples_0, leaf_samples_1 = self.shadow_tree.get_leaf_sample_counts_by_class()
 
@@ -194,8 +199,18 @@ class DTreeVizAPI:
             ax.spines['right'].set_visible(False)
             ax.spines['left'].set_linewidth(.3)
             ax.spines['bottom'].set_linewidth(.3)
-            ax.set_xticks(range(0, len(index)))
-            ax.set_xticklabels(index)
+            ax.tick_params(axis='x', which='major', labelcolor=colors['tick_label'], top=False, bottom=label_all_leaves)
+            ax.tick_params(axis='y', which='both', labelcolor=colors['tick_label'], left=True, right=False)
+            ax.tick_params(axis='x', which='minor', labelcolor=colors['tick_label'], top=False, bottom=False)
+
+            xlabel='leaf ids'
+            if label_all_leaves:
+                ax.set_xticks(range(0, len(index)))
+                ax.set_xticklabels(index)
+            else:
+                ax.set_xticklabels([])
+                xlabel='leaves'
+
             if plot_ylim is not None:
                 ax.set_ylim(0, plot_ylim)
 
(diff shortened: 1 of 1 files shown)


dtreeviz/trees.py
```diff
@@ -148,7 +148,8 @@ def ctree_leaf_distributions(self,
                                  fontname: str = "Arial",
                                  grid: bool = False,
                                  figsize: tuple = None,
-                                 ax=None):
+                                 ax=None,
```
parrt, line 151: It's weird that your PR is pulling these in because this is from one of our earlier changes. Maybe you created this PR  or branch before pulling the latest from dev?

dtreeviz/trees.py
```diff
@@ -176,8 +177,12 @@ def ctree_leaf_distributions(self,
             Plot labels font name
         :param grid: bool
             True if we want to display the grid lines on the visualization
-        :param figsize: optional (width, height) in inches for the entire plot
-            :param ax: optional matplotlib "axes" to draw into
+        :param figsize: optional
+            (width, height) in inches for the entire plot
+        :param ax: optional
+            matplotlib "axes" to draw into
+        :param label_all_leaves: bool
```
parrt, line 184: I think I have standardized figsize, ax to be the final two parameters so perhaps move yours up?

dtreeviz/trees.py
mepland: I'm not sure what you mean, but I noticed that the param doc text for figsize and ax did not have the correct format, so I fixed that at the same time.

I will move `label_all_leaves` up!

parrt:
If we decide this is the right approach, it seems like we need it on all of the leaf information functions to be consistent:

* leaf_sizes
* leaf_purity
* rtree_leaf_distributions
* ctree_leaf_distributions

mepland:
Yeah, that's a good point. I just find it useful to turn off the x-axis ticks if you are plotting a deep tree with lots of nodes. In that case, having each node getting its own tick and ticklabel makes the x-axis a mess of overlapped text.

Ideally these leaf info functions would automatically handle > ~20 nodes by making some node ids into major ticks and ignoring the rest. However, I'm not sure how to do this cleanly with the current approach of:
```
ax.set_xticks(range(0, len(index)))
ax.set_xticklabels(index)
```
We could code up our own logic around how to pick the major / minor node ids, but it would be nice to leverage any built-in `matplotlib` functionality.

dtreeviz/trees.py
parrt: oh, now it is looking clean. weird. Maybe I needed to do a refresh of the webpage.

parrt:
We can maybe clean that up later.

dtreeviz/trees.py
```diff
@@ -176,8 +177,12 @@ def ctree_leaf_distributions(self,
             Plot labels font name
         :param grid: bool
             True if we want to display the grid lines on the visualization
-        :param figsize: optional (width, height) in inches for the entire plot
-            :param ax: optional matplotlib "axes" to draw into
+        :param label_all_leaves: bool
```
parrt, line 180: I wonder if it's better to call it `show_leaf_labels`. I think we use *show* a lot.

dtreeviz/trees.py
mepland: I did a rebase to the current dev branch with that last commit.

dtreeviz/trees.py
mepland: @parrt how would you like to procede from here? Should I change the param name and we'll merge this PR, or do we want to table this and fix it in a more elegant way?

parrt: @tlapusan when you have time, maybe you have thoughts on this as you built those functions/plots?

tlapusan:
@mepland indeed, when the tree has a lot of leaves, the plot is not very nicely displayed.

@parrt I do remember that we discuss recently a similar issue and one possible solution was to either make a vertical plot or to plot the labels at a 45 degree.

If we look at the leaf_sizes() viz, we can see that we have the min_size and max_size arguments, which can help us to filter out the leaves which we are not interested. Would help to have it implemented for ctree_leaf_distributions.

@mepland to label only the major ticks I think this sample code can help, at least I tried it and put the tick and label on leaf 1 and 4.
> ax.set_xticks([1,4])
> ax.set_xticklabels(["leaf 1", "leaf 4"])

dtreeviz/trees.py
tlapusan: > I wonder if it's better to call it `show_leaf_labels`. I think we use _show_ a lot.

Agree with this. We have arguments for other methods which start with show_...

mepland: I changed it `label_all_leaves` -> `show_leaf_labels`

mepland:
> @parrt I do remember that we discuss recently a similar issue and one possible solution was to either make a vertical plot or to plot the labels at a 45 degree.

I don't think this is a sustainable solution, there will always be a bigger tree with more nodes!

> If we look at the leaf_sizes() viz, we can see that we have the min_size and max_size arguments, which can help us to filter out the leaves which we are not interested. Would help to have it implemented for ctree_leaf_distributions.

While a useful addition, I also don't think this is a full solution as users may want to have one plot with all the leaves.
 
> @mepland to label only the major ticks I think this sample code can help, at least I tried it and put the tick and label on leaf 1 and 4.
> 
> > ax.set_xticks([1,4])
> > ax.set_xticklabels(["leaf 1", "leaf 4"])

I can try something like this, it was just simpler to turn off the IDs to start :smile:

> @mepland do you have an example of a messy plot ? would help to work on it and to see how the changes will reflect in display.

Any significantly deep tree, here `DecisionTreeClassifier(max_depth=6)`, will break it. See [here](https://github.com/parrt/dtreeviz/files/10348207/ctree_leaf_distributions.pdf) for an example.

parrt:
That example makes it look like the leaf ideas actually aren't that useful. Are people really asking for leaves by id?

mepland:
It could be useful I suppose, though I am fine turning them off. I just think we should have a better solution in place to handle this mess of labels.

mepland:
@parrt where should we go from here? Should I make an issue and start a new PR that tries to do this more automatically, or are we happy with the brute force parameter that turns on/off all of the tick labels?

tlapusan:
@parrt the leaf id would be important to know when the viz will contain an 'interesting' samples distribution. In our titanic dataset would be a leaf where majority of people survived. Having the id of the leaf, I could use the note_stats(node_id) to get all the samples from that leaf and to investigate them.

@mepland when the tree will be very big and with a lot of leaves, even displaying without the leaf id will be a challange for the viz. Let's don't forget that the user has the possibility to specify the figsize, which will help. In general, when the tree will have a lot of nodes, dtreeviz will have challenges to present them nicely into a single plot.

Screenshot 2023-01-09 at 08 25 00

btw, I agree that we can have a parameter to turn off the leaf ids if the user wants to.

@mepland have you thought to other ideas when the number of leaves is large ?

mepland:
@tlapusan good point just making the plot bigger with `figsize`.

We could also plot the leaves on a "normal" axis and let matplotlib take care of the major / minor tick scaling to fit the available space. There will be potential gaps (in your plot between 5 and 10, 65 and 70, etc) if the node_ids are not sequential - but that should be fine.

Perhaps we have the current behavior, no labels, and my proposed normal matplotlib axis behavior as options?

parrt:
> Having the id of the leaf, I could use the note_stats(node_id) to get all the samples from that leaf and to investigate them.

Good point. Yes, and they can simply make the figure wider when there are lots of IDs.

@tlapusan what do you think of allowing the plotting library to decide where the ticks go depending on the figure size? Might make it hard to identify an individual node...

I'd also like to avoid an option where we can display them and not display them. I think we should always display and then let them change the figure size. Also we need to decide whether each node gets an individual label or we let the Library decide.

tlapusan:
When I proposed this viz, I had the usecase in my mind that some leaves could contain some interesting labels -> investigate them (using the leafid and node_stats(id)) -> improve domain knowledge about the dataset which could help to create new features.

Indeed, there could be another usecase when the user wants to have a general overview how the leaf distributions looks. In this case I think the leaf ids are not important and the user could have the option of not displaying them (especially when we have a lot of leaves and there is a mess for leaf labels).

About the implementing our own logic of choosing what leaf ids to present...I think that is hard to implement because there could be an infinity combinations of what a user think it's important for him to see. Better to let the user to choose what he wants to see, otherwise could be frustrating for him (at least for me :)) ).

That's why I still think that using two parameters (see the leaf_size viz) like min_size (or a better name could be min_sample) and max_size could help. I have in mind this scenario: the tree has a lot of leaves, but I can see from the viz (general overview) that some leaves have a lot of samples. In this case, I could set the min_size to a value to force the viz to show me only the leaves on which I'm interested and afterword to take some leaf ids and investigate them.

Bellow is an example for leaf_size() which could be applied to ctree_leaf_distributions() also:

Screenshot 2023-01-15 at 18 54 47

parrt:
Thanks for the explanation. How about something more general and instead of `min_size`, we pass in a function or lambda that takes a size parameter and returns to or false as to whether it should be included?

Then, we need an option to turn off/on the leaf ids.

Does that make sense? Seems we would then make an analogous change for the other leaf stats methods, right?

mepland:
I'm going to close this PR for now, let's make a new branch after 2.1 is released and we decide what changes we'd like to make.

mepland:
Started an issue to track: https://github.com/parrt/dtreeviz/issues/247

PR #223, opened 2022-12-29
Fix default larrow rarrow

mepland:
@parrt Let's leave `larrow, rarrow` in `colors.py` so users see they exist, but comment them out like this. The default None does not print any edge arrows.


dtreeviz/colors.py
```diff
@@ -95,8 +95,8 @@ def get_hex_colors(n_classes, cmap_name="RdYlBu"):
           'wedge': WEDGE_COLOR,
           'text_wedge': WEDGE_COLOR,
           'arrow': GREY,
-          'larrow': None,
-          'rarrow': None,
+          # 'larrow': GREY,
```
parrt, line 98: hmm...commented out code is usually "wrong". 

> The default None does not print any edge arrows.

Maybe we change this

```
if 'larrow' in colors and 'rarrow' in colors:
```

so we check it's value not presence?  in python we likely need both to avoid exception.

mepland: How about this?

parrt: hiya. I still don't think I like the commented out code stuff. It looks like a typo as you have it now. the `None` makes it clear.  Hmm... Actually I think here is the simple solution... Leave them as None but change logic to:

```
                lcolor = colors.get('larrow', colors['arrow'])
                rcolor = colors.get('rarrow', colors['arrow'])
```

It even gets rid of the need for the specific exception. It also uses the `arrow` as the back up color, allowing you to specify just one, left or right.

mepland: You might not have refreshed. I wrote a different `.get()` solution, but yours is better - with the fall back to `arrow`. Do you want to implement it, or me?

Also, I don't think most of the color dict keys are checked / error handled in general.

parrt: You might as well updated since you have the PR. 

> Also, I don't think most of the color dict keys are checked / error handled in general.

Ah. yeah, true. I guess in most cases we are expecting people not to blast values in the dictionary.

mepland: Done!
dtreeviz

parrt:
thanks for all the help!

PR #224, opened 2022-12-30
Fix Digraph Graphics

mepland:
- Changed additional hard coded `Helvetica` to `fontname` in digraph.
- Added `fontname="{fontname}"` for digraph edges, now all fonts should be the same.
- Added leading spaces to `all_llabel, all_rlabel` and `root_llabel, root_rlabel`.
- Noted that the `" &#160;"` in `edge_label` is not doing anything for pdf views, see attached.

dtreeviz_pred_path
[dtreeviz_pred_path.pdf](https://github.com/parrt/dtreeviz/files/10322602/dtreeviz_pred_path.pdf)

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -432,10 +432,9 @@ class DTreeVizAPI:
             path = self.shadow_tree.predict_path(x)
             leaf = f"leaf{path[-1].id}"
             if self.shadow_tree.is_classifier():
-                # TODO both regular spaces, &nbsp;, and &#160; do not move edge_label away from the line
-                edge_label = f" &#160;Prediction<br/> {path[-1].prediction_name()}"
+                edge_label = f"  Prediction<br/> {path[-1].prediction_name()}"
             else:
-                edge_label = f" &#160;Prediction<br/> {myround(path[-1].prediction(), precision)}"
+                edge_label = f"  Prediction<br/> {myround(path[-1].prediction(), precision)}"
             return f"""
                     subgraph cluster_instance {{
                         style=invis;
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -347,7 +347,7 @@ class DTreeVizAPI:
                 return f'leaf{node.id} [margin="0" shape=box penwidth="0" color="{colors["text"]}" label=<{html}>]'
 
         def node_label(node):
-            return f'<tr><td CELLPADDING="0" CELLSPACING="0"><font face="Helvetica" color="{colors["node_label"]}" point-size="14"><i>Node {node.id}</i></font></td></tr>'
+            return f'<tr><td CELLPADDING="0" CELLSPACING="0"><font face="{fontname}" color="{colors["node_label"]}" point-size="14"><i>Node {node.id}</i></font></td></tr>'
 
         def class_legend_html():
             return f"""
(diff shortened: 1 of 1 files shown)


mepland:
Hmm actually looking at it again, the "Prediction 0" text is in the correct spot on the png I save, but not the pdf, so I think that it is fine. I just changed to two leading spaces to match the other labels. The `&#160;` [character](https://stackoverflow.com/questions/3274315/is-160-a-replacement-of-nbsp) doesn't appear to be doing anything in my testing.

Ideally we would just move the edge label positions in the graph itself, but this doesn't appear possible for graphviz:

- [Ref 1](https://stackoverflow.com/questions/67854248/set-distance-between-arrow-and-label)
- [Ref 2](https://stackoverflow.com/questions/3450700/graphviz-place-edge-label-on-the-other-side)
- [Ref 3](https://stackoverflow.com/questions/18515529/graphviz-place-edge-label-on-the-other-side-ii)
- [Ref 4](https://graphviz.org/doc/info/attrs.html#d:labeldistance) - Note `labeldistance` only works for `headlabel` and `taillabel`, not the midpoint `label` we use.

dtreeviz/trees.py
```diff
@@ -585,10 +585,10 @@ def get_leaves():
 
         # TODO do we need show_edge_labels ?
         show_edge_labels = False
-        all_llabel = '&lt;' if show_edge_labels else ''
-        all_rlabel = '&ge;' if show_edge_labels else ''
-        root_llabel = self.shadow_tree.get_root_edge_labels()[0] if show_root_edge_labels else ''
-        root_rlabel = self.shadow_tree.get_root_edge_labels()[1] if show_root_edge_labels else ''
+        all_llabel = '  &lt;' if show_edge_labels else ''
```
parrt, line 588: Do we have a before and after picture the demonstrates why the extra space is needed?

dtreeviz/trees.py
```diff
@@ -629,12 +629,12 @@ def get_leaves():
...
-                    edges.append(f'{nname} -> {left_node_name} [penwidth={lpw} color="{lcolor}" label=<{llabel}> fontcolor="{colors["text"]}"]')
+                    edges.append(f'{nname} -> {left_node_name} [penwidth={lpw} fontname="{fontname}" color="{lcolor}" label=<{llabel}> fontcolor="{colors["text"]}"]')
                 if node.right.id in highlight_path:
-                    edges.append(f'{nname} -> {right_node_name} [penwidth={rpw} color="{rcolor}" label=<{rlabel}> fontcolor="{colors["text"]}"]')
+                    edges.append(f'{nname} -> {right_node_name} [penwidth={rpw} fontname="{fontname}" color="{rcolor}" label=<{rlabel}> fontcolor="{colors["text"]}"]')
             else:
-                edges.append(f'{nname} -> {left_node_name} [penwidth={lpw} color="{lcolor}" label=<{llabel}> fontcolor="{colors["text"]}"]')
-                edges.append(f'{nname} -> {right_node_name} [penwidth={rpw} color="{rcolor}" label=<{rlabel}> fontcolor="{colors["text"]}"]')
+                edges.append(f'{nname} -> {left_node_name} [penwidth={lpw} fontname="{fontname}" color="{lcolor}" label=<{llabel}> fontcolor="{colors["text"]}"]')
```
parrt, line 636: Awesome. Thanks for cleaning all of this fontname stuff up.

parrt:
> Ideally we would just move the edge label positions in the graph itself, but this doesn't appear possible for graphviz:

yeah, I had lots of trouble with positioning...

dtreeviz/trees.py
mepland: I don't have a before an after at the ready, but when the tree is large / wide and the edge arrows are spread apart, the inequality symbols can overlap the arrows. This gives some extra room to help avoid that.

parrt:
just ran notebooks. looks fine. thanks!

PR #227, opened 2022-12-30
Improve _ctreeviz_univar()

mepland:
- Compute bins only when needed
- Fix crash for 0 count bins
- Fix split heights changing in for loop
- Make new `'preds'` key in show param to separately control the horizontal prediction bars
- Add `ValueError` for unrecognized `gtype`
- Match other rectangle `rect_edge` color style

image


mepland:
- Added tesselation_alpha to let user make horizontal 'preds' bars transparent if they wish
image

dtreeviz/trees.py
```diff
@@ -1492,23 +1492,27 @@ def _ctreeviz_univar(shadow_tree,
...
 
     ax.tick_params(axis='both', which='major', width=.3, labelcolor=colors['tick_label'],
                    labelsize=fontsize)
 
     splits = [node.split() for node in shadow_tree.internal]
     splits = sorted(splits)
-    bins = [ax.get_xlim()[0]] + splits + [ax.get_xlim()[1]]
 
-    if 'splits' in show:  # this gets the horiz bars showing prediction region
+    if 'preds' in show:  # this gets the horiz bars showing prediction region
```
parrt, line 1504: So this looks like it's renaming rather than adding functionality?

parrt: OK I get it now. You are separating the dashed vertical lines from the prediction horizontal boxes.

parrt:
The prediction bars look like they are overlapping the bottom of the bar charts. Is this the previous behavior and I just didn't notice? dang. This PR looks fine, but let's leave this open until we can resolve this issue I just raised. Seems like the prediction should go below the axis not on top of the others.

dtreeviz/trees.py
```diff
@@ -1506,6 +1506,8 @@ def _ctreeviz_univar(shadow_tree,
             left = bins[i]
             right = bins[i + 1]
             inrange = y_train[(X_train >= left) & (X_train <= right)]
+            if 0 == len(inrange):
```
parrt, line 1511: I think I prefer the operations in reverse of this order :)

dtreeviz/trees.py
```diff
@@ -1526,8 +1526,9 @@ def _ctreeviz_univar(shadow_tree,
         ax.set_title(title, fontsize=fontsize, color=colors['title'])
 
     if 'splits' in show:
+        split_heights = [*ax.get_ylim()]
```
parrt, line 1529: I didn't realize these could change height... wow. i guess the plot could do that.

dtreeviz/trees.py
mepland: Yeah for 2 reasons.

1. As you note below, the prediction horizontal boxes overlap the bottom of the bars.
2. It would be nice to turn the prediction boxes off when plotting the splits of a larger (more than 1-2 feature) tree, as the prediction regions could overlap.

parrt: predictions in a uni-var feature space shouldn't have any overlap, should there?  Every rectangular region inscribed by a decision tree's leaf node(s) will have a unique prediction I think.

mepland:
> The prediction bars look like they are overlapping the bottom of the bar charts. Is this the previous behavior and I just didn't notice? dang. This PR looks fine, but let's leave this open until we can resolve this issue I just raised. Seems like the prediction should go below the axis not on top of the others.

@parrt https://github.com/parrt/dtreeviz/pull/227/commits/3a72aa21753659abb83aef4770e4aa9357cd1d12 moves the `'preds'` boxes below the x-axis. What do you think?

image

parrt:
> What do you think?

looks great!

mepland:
@parrt I think we should remove the [tesselation_alpha](https://github.com/parrt/dtreeviz/pull/227#issuecomment-1367699248) parameter, now that overlaps are impossible. I can do it in a new branch and PR.

parrt:
That sounds right. Thank you

mepland:
> That sounds right. Thank you

Done, see https://github.com/parrt/dtreeviz/pull/235

PR #230, opened 2022-12-30
Fix spelling of tessellation

parrt:


parrt:
I think this will only affect the people using the classifier boundary view so I will merge; not sure how many people are using that, but now it's time to clean up.

PR #232, opened 2023-01-03
Refactored Axes Formatting

mepland:
@parrt I refactored the axes formatting code. The plots should be much the same, but more consistent and with less code in `trees.py`.


mepland:
I also added a new `'axis'` color parameter to adjust the axis spline color.

parrt:
I will wait to hear from @tlapusan on this one.

tlapusan:
I checked both the code and the visualisations.

There are some slightly changes in the viz, but in a good way, making them more consistence...this is also the role of using the same code from _format_axes().

master_version_viz
pr_version_viz


thanks @mepland for this refactoring.

Issue #231, opened 2023-01-01
Dimension issue in dtreeviz_sklearn_pipeline_visualisations.ipynb


mepland:
@parrt @tlapusan Looks like there is a bug with `extract_params_from_pipeline()`. Try running `dtreeviz_sklearn_pipeline_visualisations.ipynb` in the `dev` branch.

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[10], line 1
----> 1 tree_classifier, X_train, features_model = extract_params_from_pipeline(
      2     pipeline=model,
      3     X_train=dataset[features],
      4     feature_names=features)

File ~/dtreeviz/dtreeviz/utils.py:192, in extract_params_from_pipeline(pipeline, X_train, feature_names)
    186 tree_model = pipeline.steps[-1][1]
    188 feature_names = _extract_final_feature_names(
    189     pipeline=pipeline,
    190     features=feature_names
    191 )
--> 192 X_train = pd.DataFrame(
    193     data=pipeline[:-1].transform(X_train),
    194     columns=feature_names
    195 )
    196 return tree_model, X_train, feature_names

File ~/.venvs/dtreeviz/lib64/python3.11/site-packages/pandas/core/frame.py:721, in DataFrame.__init__(self, data, index, columns, dtype, copy)
    711         mgr = dict_to_mgr(
    712             # error: Item "ndarray" of "Union[ndarray, Series, Index]" has no
    713             # attribute "name"
   (...)
    718             typ=manager,
    719         )
    720     else:
--> 721         mgr = ndarray_to_mgr(
    722             data,
    723             index,
    724             columns,
    725             dtype=dtype,
    726             copy=copy,
    727             typ=manager,
    728         )
    730 # For data is list-like, or Iterable (will consume into list)
    731 elif is_list_like(data):

File ~/.venvs/dtreeviz/lib64/python3.11/site-packages/pandas/core/internals/construction.py:349, in ndarray_to_mgr(values, index, columns, dtype, copy, typ)
    344 # _prep_ndarraylike ensures that values.ndim == 2 at this point
    345 index, columns = _get_axes(
    346     values.shape[0], values.shape[1], index=index, columns=columns
    347 )
--> 349 _check_values_indices_shape_match(values, index, columns)
    351 if typ == "array":
    353     if issubclass(values.dtype.type, str):

[...]

PR #233, opened 2023-01-03
Adapt utility function for `sklearn>=1.2.0`

windisch:
This addresses #231 and refactors the `setup.py` a bit (open for discussion).

The feature extraction from sklearn-pipelines becomes much easier in version `1.2.0`. I added a if-condition (similar as suggested by @tlapusan) that checks whether the pipeline allows for this new feature. However, in the long run, maintaining compatibility to all the sklearn versions will increase the code complexity. I recommend to add a lower-bound on the `sklearn` version at some point.

### DIFF
--- a/dtreeviz/utils.py
+++ b/dtreeviz/utils.py
@@ -145,13 +145,24 @@ def _extract_final_feature_names(pipeline, features):
 
     Returns:
         list: Features names used by the last component
+
+    Note:
+        This function depends on how the feature names in :py:mod:`sklearn` are handled. Any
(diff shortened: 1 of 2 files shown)


tlapusan:
@windisch thanks for taking time and making these changes.

I've tested your changes related to sklearn and they work. Indeed, any version change for any library could break the code, especially for major releases... but we hope we are gonna patch them :)

I'm ok also with the setup.py refactor, do you see any benefit of creating those variable for extra_* ?

Issue #204, opened 2022-12-10
Prediction path explanation should show comparisons in order of node discovery


parrt:
Currently the output for the prediction path is given in the feature order in the data frame.

mepland:
Do we want to only display the features used in the prediction? I think the user can print a dataframe row if they want all of the features, but including them is making the text box large, or adding the ellipsis.

![dtreeviz_pred_path](https://user-images.githubusercontent.com/4729931/210300006-df130d93-02d1-4440-aefd-4c7e396ae8be.png)

tlapusan:
@mepland I think that @parrt is referring to this viz 
<img width="538" alt="Screenshot 2023-01-03 at 09 58 47" src="https://user-images.githubusercontent.com/12815158/210318635-9885d8d8-778a-4efe-9e7c-d278939a67eb.png">

mepland:
@parrt @tlapusan done, see https://github.com/parrt/dtreeviz/pull/234

parrt:
Regarding showing the entire instance of X, probably best to show at all but I can see a case where it's too big.

parrt:
Closing per our discussion https://github.com/parrt/dtreeviz/pull/234#discussion_r1064108697

PR #234, opened 2023-01-04
Order explain_prediction_path()

mepland:
Save distinct feature_name values as we traverse the tree, so we can print them in the same order later. Addresses https://github.com/parrt/dtreeviz/issues/204

image

### DIFF
--- a/dtreeviz/interpretation.py
+++ b/dtreeviz/interpretation.py
@@ -44,6 +44,7 @@ def explain_prediction_plain_english(shadow_tree: ShadowDecTree,
     feature_categorical_value = defaultdict(lambda: set())
     feature_categorical_value_not_in = defaultdict(lambda: set())
 
+    decision_node_path_ordered_feature_names = []
     for i, node in enumerate(decision_node_path):
         if i == len(decision_node_path) - 1:
             break  # stop at leaf node
@@ -51,6 +52,9 @@ def explain_prediction_plain_english(shadow_tree: ShadowDecTree,
 
         feature_name = feature_names[node_feature_index[node_id]]
         feature_value = x[node_feature_index[node_id]]
+
+        if feature_name not in decision_node_path_ordered_feature_names:
+            decision_node_path_ordered_feature_names.append(feature_name)
 
         if not shadow_tree.is_categorical_split(node_id):
             feature_split_value = round(node_threshold[node_id], 2)
@@ -70,7 +74,7 @@ def explain_prediction_plain_english(shadow_tree: ShadowDecTree,
                 feature_categorical_value_not_in[feature_name].update(node_threshold[node_id])
 
     prediction_path_output = ""
-    for feature_name in feature_names:
+    for feature_name in decision_node_path_ordered_feature_names:
         feature_range = ""
         if feature_name in feature_smaller_values:
             feature_range = f"{max(feature_smaller_values[feature_name])} <= {feature_name} "


dtreeviz/interpretation.py
```diff
@@ -52,6 +53,9 @@ def explain_prediction_plain_english(shadow_tree: ShadowDecTree,
         feature_name = feature_names[node_feature_index[node_id]]
         feature_value = x[node_feature_index[node_id]]
 
+        if feature_name not in decision_node_path_ordered_feature_names:
```
parrt, line 56: I'm wondering if this is right. How does this handle a tree that tests the same variable multiple times?

mepland: The first time the that `feature_name` appears it is added to the list, if it's already been seen - and thus is in the list - it is ignored / not added a second time. So it will not appear more than once, but the exact order will depend on how the tree is traversed from top to bottom.

parrt: Hmm... if tree tests x_i 8 times it should have 8 lines in the output right?

mepland: @parrt The prior code was already only printing each feature one time, no matter how many times it appeared in the decision path. The `feature_names = shadow_tree.feature_names` list only has one copy of each `feature_name`, right?.

I assumed the rest of the `explain_prediction_plain_english()` logic, under the `# TODO - refactor this logic and find a way to make it simpler` note, would make the correct inequality that covered all cases for a feature, e.g. if x_1 < 1 AND x_1 < 2 AND x_1 < 3 are in the decision path, then we'd only print x_1 < 3. Is this not the case?

parrt: Oh that is possible. I don't see any join with `AND` though from a casual inspection. We better wait to hear from @tlapusan 

I guess it does make the most sense that all tests of a single variable would be grouped together on a single line. If we go that route, then it doesn't really make sense to order things by the path through the tree anymore as I was suggesting. Instead, it would be better to order as Tudor had it before: by column order in X.

tlapusan: indeed, the current logic groups together all the feature split values in a single line. 

@mepland yes, I put that note because I do feel that logic is not in the best shape and is hard to understand :)) 

thinking now, yes, maybe is no value added to present them by the path order. Would be interesting to order them by prediction or tree feature importance.

parrt: OK so it sounds like we are all agreed. Let's group all references to the same variable on the same line and do them in the order specified in the X columns. It would be nice if they were properly formatted as AND and ordered smallest to greatest value if numerical?

mepland: That's not what I was thinking, but sounds fine. @parrt can we abandon this PR then, close #204, and open a new issue with the exact specs you'd like to see for the output?

Something like this, when including both `<` and `>`:
```
x_1 = 0 < 1
2 < x_3 = 3 AND x_3 = 3 < 5 AND x_3 = 3 < 10
x_11 = 11 < 20
```

Rows are ordered by `x_i`. Within a row, all splits on `x_i` are shown, from smallest to largest, connected with `AND`?

parrt: sounds good to me.

Do all trees treat categoricals with < and > not "in set" and "not in set"?

tlapusan: > That's not what I was thinking, but sounds fine. @parrt can we abandon this PR then, close #204, and open a new issue with the exact specs you'd like to see for the output?
> 
> Something like this, when including both `<` and `>`:
> 
> ```
> x_1 = 0 < 1
> 2 < x_3 = 3 AND x_3 = 3 < 5 AND x_3 = 3 < 10
> x_11 = 11 < 20
> ```
> 
> Rows are ordered by `x_i`. Within a row, all splits on `x_i` are shown, from smallest to largest, connected with `AND`?

@mepland 
The curent logic will output for the x_3 feature something like this:
2 <= x_3 < 5
because 5 is smaller than 10, we don't have to include 10 in explanation. The idea is to find the most relevant split value range. Imagine if the prediction path will contain many split value for the same feature...very hard to understand/compact the split value range.

@parrt yes, when it's about a categorical feature, the explanation will contain "in" or "not in", like
Screenshot 2023-01-08 at 11 16 09

I do remember that I needed to include the "not in" because this was the way Spark handle the categorical split checks, feature_x "not in" {set of values} vs feature_x in {set of values}

mepland: @tlapusan @parrt I'm fine with whatever for `explain_prediction_plain_english()`. Can we close this PR and I'll let you two do any updates you'd like?

parrt: OK so it sounds like some models are able to handle categoricals where is I think with sklearn, it's always just numeric values. Closing this one in favor of creating an issue which we can fix later https://github.com/parrt/dtreeviz/issues/240

PR #236, opened 2023-01-05
Update Wedges

mepland:


mepland:
@parrt this branch / PR now has a working version of the wedges code refactored into a utility function. However, I have yet to address the overlapping text. A few issues:

- When we put the orange prediction path wedges on the graph they can also overlap an x-axis tick. This is probably more likely than the black split wedge overlapping, as a single data point is more likely to be at the min/max of the axis than the split itself.
image

- The regression trees are creating an extra x-axis tick under the split and labeling it, while the classification trees are adding that extra text annotation. Adding an extra axis tick may be cleaner, it will have the same font and position as the min/max axis ticks. We should decide what way we prefer and standardize between the two. In this plot you can see both labels are being drawn, they are slightly different:
image

- The new categorical feature support for the classification tree can add even more split wedges, increasing the possibility of overlapping graphics:
```
            # classification: categorical split, draw multiple wedges, no labels
            bins = np.linspace(start=overall_feature_range[0], stop=overall_feature_range[1], num=nbins, endpoint=True)
            for split_value in node.split():
                # to display the wedge exactly in the middle of the vertical bar
                for bin_index in range(len(bins) - 1):
                    if bins[bin_index] <= split_value <= bins[bin_index + 1]:
                        split_value = (bins[bin_index] + bins[bin_index + 1]) / 2
                        break
                _draw_tria(split_value, tip_y, tri_width, tri_height)
```


I think the best path forward maybe to:

1. Add extra ticks to the x-axis where the wedges fall, if we want to label them with a numeric value, i.e. splits but not prediction path wedges. Basically, standardize on what the regression tree plots are doing currently.
2. Move the x-axis tick labels further away from the axis so the wedges never overlap them, e.g. `ax.tick_params(axis='x', pad=30)`. This would also keep all the plots the uniform, regardless of if they have a wedge or not.
3. Write code to take the list of x-axis ticks and drop ticks if the labels would overlap. Basically, keep the min and max tick, then go from left to right, keeping the tick if it is not within some (10%) distance of it's neighbors.

Let me know what you think and I can code it up, but it will not be ready for tomorrows 2.1 release.

parrt:
I like the sound of that approach but until we can see what it looks like with the wedges separated from the X axis, it's hard to tell whether it's right or not. Are you OK making an experiment add mermaid not be exactly what we use?

mepland:
@parrt Yeah I can code up a demo, even if it is not fully implemented. I'd like to wait till after https://github.com/parrt/dtreeviz/pull/232 is merged though, as it pertains to the axis formatting.

mepland:
@parrt I have the wedge code refactored and have fixed the overlapping issue!

Basically, the `_draw_wedge()` function draws the wedges in a common way, returning a list of their location(s). Then `_set_wedge_ticks()` deconflicts the wedge ticks from the desired axis ticks, giving priority to the axis ticks. For the plots with potential wedges we always move the tick labels down with `pad=8` so the tick labels and wedges never overlap and all the plots are consistent.

Some examples:

- The wedges are unlabeled / not ticked if they are within 10% of the min / max ticks, classification trees
image

- Same behavior for regression trees
image

- Prediction path ticks are currently never labeled, as that was the previous behavior. I can change this if we want. It will require more deconfliction between the prediction path wedge ticks and split wedge ticks, but is doable. I can probably even set the [predicted path wedge tick labels to have the same color](https://stackoverflow.com/a/42879231). @parrt please let me know if you'd like this feature and I can add it, otherwise the PR should be ready to merge!
image

parrt:
Looks great! I think not putting the x_i value for the prediction path orange wedges sounds right because it declutters and we know the value because we can see the X vector below.

parrt:
I'm not sure what the pad_for_wedge does as it is always true. I'm trying to look at the difference between the images. How does the padding help us?

> For the plots with potential wedges we always move the tick labels down with pad=8 so the tick labels and wedges never overlap and all the plots are consistent.

Do you mean for prediction path wedges when you say potential wedges?

Sounds like we should just make this always happen or probably I'm missing calls that have this use the default value of false?

mepland:
> I'm not sure what the pad_for_wedge does as it is always true. I'm trying to look at the difference between the images. How does the padding help us?

`pad_for_wedge` is only true for the split plots were we are drawing ledges, in these two calls of `_format_axes()`:
```
_format_axes(ax, feature_name, None, colors, fontsize=label_fontsize, fontname=fontname, ticks_fontsize=ticks_fontsize, grid=False, pad_for_wedge=True)
_format_axes(ax, feature_name, target_name if node == node.shadow_tree.root else None, colors, fontsize=label_fontsize, fontname=fontname, ticks_fontsize=ticks_fontsize, grid=False, pad_for_wedge=True)
```
 It is false by default in all the other `_format_axes()` calls were we don't have wedges and want the x-axis ticks to have matplotlib's default padding.

What this parameter does is increase the spacing between the x-axis tick labels and the axis line itself, so the wedges and text will never overlap.
```
if pad_for_wedge:
    ax.tick_params(axis='x', which='major', pad=8)
```

Notice how the 0.42 here is actually labeling the x-axis min tick, but due to the extra padding it doesn't overlap the split wedge vertically. Also in this particular example the split did not get a tick as it was with in 10% of the existing tick.
image

mepland:
@parrt are we good to merge this PR then?

parrt:
yep. Distracted working on the tutorial for tensor flow decision trees... I have to push out a Version soon so that I can import it into that notebook; can't test locally due to crashes on my M1 Mac.

parrt:
Looks like this has conflict. Can you resolve?

mepland:
@parrt there should be no conflict?

parrt:
Screen Shot 2023-01-14 at 2 35 43 PM

mepland:
@parrt looks good to go on my end, and I also just merged `dev` into `update_wedges` so if it still shows a conflict for you I'm not sure what to do...
image

parrt:
yeah, seen this before...booo. i'll switch machines and merge.  oh, rebase doesn't work but squash/merge does.

PR #245, opened 2023-01-15
Capitalization Fixes

mepland:
Also refactored missed `_format_axes()` in `interpretation.py` and `classifiers.py`

### DIFF
--- a/dtreeviz/models/sklearn_decision_trees.py
+++ b/dtreeviz/models/sklearn_decision_trees.py
@@ -5,6 +5,7 @@ import numpy as np
 from sklearn.utils import compute_class_weight
 
 from dtreeviz.models.shadow_decision_tree import ShadowDecTree
+from dtreeviz.utils import criterion_remapping
 
 
 class ShadowSKDTree(ShadowDecTree):
@@ -36,7 +37,7 @@ class ShadowSKDTree(ShadowDecTree):
         return self.tree_model.tree_.feature
 
     def criterion(self):
-        return self.tree_model.criterion
+        return criterion_remapping.get(self.tree_model.criterion, self.tree_model.criterion)
 
     def get_class_weight(self):
         return self.tree_model.class_weight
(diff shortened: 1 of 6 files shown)


mepland:
@parrt I don't know why these lines are included to change the x-axis ticks of the importance plots, but I think they should be removed.

Before:
```
    # TODO this just screws up the x-axis ticks, should just leave numeric!
    ax.set_xticks(range(0, len(shadow_tree.feature_names)))
    ax.set_xticklabels(shadow_tree.feature_names)
```
image

After (https://github.com/parrt/dtreeviz/pull/245/commits/73462af699cfa0f8dc6542af5e7a847d72488287):
image

parrt:
> I don't know why these lines are included to change the x-axis ticks of the importance plots, but I think they should be removed.

Agreed. I believe I recently change this to be horizontal and probably left some code which is strange but yes all of the feature names should be on the vertical axis now.

parrt:
> I looked through the documentation for sklearn and spark trees and pulled all of the criterion values I could find.

Excellent

dtreeviz/models/sklearn_decision_trees.py
```diff
@@ -36,7 +37,7 @@ def get_features(self):
         return self.tree_model.tree_.feature
 
     def criterion(self):
-        return self.tree_model.criterion.upper()
+        return criterion_remapping.get(self.tree_model.criterion, self.tree_model.criterion)
```
parrt, line 40: I wonder if we should just make a function that automatically gets the default value to hide the double reference to the criterion in the function call.

dtreeviz/models/spark_decision_tree.py
```diff
@@ -99,7 +100,7 @@ def get_features(self) -> np.ndarray:
         return self.features
 
     def criterion(self) -> str:
-        return self._get_tree_model_parameter_value("impurity")
+        return criterion_remapping.get(self._get_tree_model_parameter_value("impurity"), self._get_tree_model_parameter_value("impurity"))
```
parrt, line 103: A small issue but might be better to simplify this into a function call. As above.

dtreeviz/models/sklearn_decision_trees.py
mepland: Done! https://github.com/parrt/dtreeviz/pull/245/commits/c0f6a0fe6b620bd85a4a8a3f0fd61c68271e866b

PR #253, opened 2023-01-21
Got basics of new ctree_feature_space() that accepts a features arg.

parrt:
Signed-off-by: Terence Parr <parrt@antlr.org>

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -1447,7 +1447,8 @@ def _ctreeviz_bivar(shadow_tree, fontsize, ticks_fontsize, fontname, show,
         ax.scatter(h[:, featidx[0]], h[:, featidx[1]], marker='o', s=dot_w, c=color_map[i],
                    edgecolors=colors['scatter_edge'], lw=.3)
 
-    _format_axes(ax, shadow_tree.feature_names[featidx[0]], shadow_tree.feature_names[featidx[1]], colors, fontsize, fontname, ticks_fontsize=ticks_fontsize, grid=False)
+    _format_axes(ax, shadow_tree.feature_names[featidx[0]], shadow_tree.feature_names[featidx[1]],
+                 colors, fontsize, fontname, ticks_fontsize=ticks_fontsize, grid=False)
 
     if 'legend' in show:
         add_classifier_legend(ax, shadow_tree.class_names, class_values, color_map, shadow_tree.target_name, colors,
@@ -1526,7 +1527,7 @@ def _rtreeviz_bivar_heatmap(shadow_tree, fontsize, ticks_fontsize, fontname,
                             markersize,
                             features,
                             figsize,
-                            ax) -> tree.DecisionTreeClassifier:
+                            ax):
     """
     Show tesselated 2D feature space for bivariate regression tree. X_train can
     have lots of features but features lists indexes of 2 features to train tree with.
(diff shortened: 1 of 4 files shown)


parrt:
ok, ready for review! I updated sklearn notebook.

mepland:
I think we should turn the y-axis on for `univar` when `gtype='barstacked'` otherwise you can't tell anything from the ' magnitude.

image

parrt:
I guess it’s just a count, right?

tlapusan:
Is there a need to reinitialize the decision tree/dtreeviz in case we want to use one and after that two features ?
I think what @mepland initially suggested is to initialize a decision tree/dtreeviz model and after that to choose if you want to make a viz with one or two features. Like in the bellow screenshot :

Screenshot 2023-01-22 at 15 47 37

mepland:
> Is there a need to reinitialize the decision tree/dtreeviz in case we want to use one and after that two features ?

I think that is the current behavior, the screenshot looks fine to me?

tlapusan:
I took the code from this PR and there is reinitialization for both classification and regression... @mepland can you check this also please ?

parrt:
Good catch @tlapusan. Sorry about that. cleaned up and resubmitted the notebook so that it creates a single decision tree and model, then reuses that for visualization

mepland:
I see what you mean now @tlapusan.

@parrt looks good to me now, but I would still like the option of having a y-axis for `gtype='barstacked'`.

dtreeviz/trees.py
```diff
@@ -886,27 +886,33 @@ def ctree_feature_space(self,
...
                             figsize=None,
                             ax=None):
         # TODO: check if we can find some common functionality between univar and bivar visualisations and refactor
         #  to a single method.
-        if len(self.shadow_tree.feature_names) == 1:     # univar example
-            _ctreeviz_univar(self.shadow_tree, fontsize, ticks_fontsize, fontname, nbins, gtype, show, colors, figsize, ax)
-        elif len(self.shadow_tree.feature_names) == 2:   # bivar example
-            _ctreeviz_bivar(self.shadow_tree, fontsize, ticks_fontsize, fontname, show, colors, figsize, ax)
+        if features is None:
+            features = self.shadow_tree.feature_names[0:2] # pick first two features if none given
```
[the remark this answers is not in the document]
mepland: @parrt I forgot to submit my review, so I don't think you saw this comment.

dtreeviz/trees.py
```diff
@@ -886,27 +886,33 @@ def ctree_feature_space(self,
...
                              f" You provided a dataset with {len(self.shadow_tree.feature_names)} features {self.shadow_tree.feature_names}.")
 
     def rtree_feature_space(self, fontsize: int = 10, ticks_fontsize=8, show={'title', 'splits'}, split_linewidth=.5,
                             mean_linewidth=2, markersize=15, colors=None, fontname="Arial",
-                            n_colors_in_map=100, figsize=None, ax=None):
-        if len(self.shadow_tree.feature_names) == 1:  # univar example
+                            n_colors_in_map=100, features=None,
+                            figsize=None, ax=None):
+        if features is None:
+            features = self.shadow_tree.feature_names[0:2] # pick first two features if none given
```
mepland, line 909: Same

mepland (changes requested):

parrt:
> @parrt looks good to me now, but I would still like the option of having a y-axis for `gtype='barstacked'`.

Fixed

Screenshot 2023-01-22 at 11 15 59 AM

dtreeviz/trees.py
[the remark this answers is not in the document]
parrt: I just pushed to fix. Thanks.

parrt:
OK, can you guys do one last quick check? If it looks good I can push it for a quick release so I can use it within colab.

dtreeviz/trees.py
```diff
@@ -915,14 +923,21 @@ def rtree_feature_space3D(self,
...
+                              figsize=None, ax=None):
         """
         Show 3D feature space for bivariate regression tree. X_train should have
         just the 2 variables used for training.
         """
-        _rtreeviz_bivar_3D(self.shadow_tree, fontsize, ticks_fontsize, fontname, azim, elev, dist, show, colors,
-                           markersize,
-                           n_colors_in_map, figsize, ax)
+        if features is None:
+            features = self.shadow_tree.feature_names[0:2] # pick first two features if none given
```
mepland, line 934: @parrt do you want to add a safety check here as well, or just let it crash on the user if they have < 3 features?

parrt: oh crap, i missed one? hang on.  done.

dtreeviz/trees.py
```diff
@@ -931,7 +931,8 @@ def rtree_feature_space3D(self,
         just the 2 variables used for training.
         """
         if features is None:
-            features = self.shadow_tree.feature_names[0:2] # pick first two features if none given
+            n_features = len(self.shadow_tree.feature_names)
+            features = self.shadow_tree.feature_names[0:min(n_features,2)] # pick first one/two features if none given
```
mepland, line 935: @parrt wouldn't `_rtreeviz_bivar_3D` crash anyway if `len(features) < 3`? I think it should just raise a `ValueError`.

mepland (changes requested):

dtreeviz/trees.py
parrt: Cool. Improved to check for !=2. How'z it look?

mepland: LGTM! I cleaned up the comment in a commit.

mepland (approved):

parrt:
Cool. I will push out a new version today.

PR #259, opened 2023-02-03
handle more than 2 classes

### LINKED ISSUE #252: can't handle > 2 classes for ctree_leaf_distributions() and maybe others
```
    # TODO
    # implementation should happen in shadow tree implementations, we already have methods for this
    # this implementation will work also for validation dataset.... think how to separate model interpretation using training vs validation dataset.
    def n_sample_classes(self):
        """Used for binary classification only.

        Returns the sample count values for each classes.
        """
        samples = np.array(self.samples())
        if samples.size == 0:
            return [0, 0]

        node_y_data = self.shadow_tree.y_train[samples]
        unique, counts = np.unique(node_y_data, return_counts=True)

        if len(unique) == 2:
            return [counts[0], counts[1]]
        elif len(unique) == 1:  # one node can contain samples from only on class
            if unique[0] == 0:
                return [counts[0], 0]
            elif unique[0] == 1:
                return [0, counts[0]]
```

tlapusan:
Handle more than 2 classes for both view() and ctree_leaf_distributions()

Should close #252

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -131,7 +131,7 @@ class DTreeVizAPI:
 
 
     def ctree_leaf_distributions(self,
-                                 display_type: str = "plot",
+                                 display_type: ("plot", "text") = "plot",
                                  plot_ylim: int = None,
                                  colors: dict = None,
                                  fontsize: int = 10,
@@ -169,7 +169,6 @@ class DTreeVizAPI:
             :param ax: optional matplotlib "axes" to draw into
         """
         index, leaf_samples = self.shadow_tree.get_leaf_sample_counts_by_class()
-
         if display_type == "plot":
             colors = adjust_colors(colors)
             colors_classes = colors['classes'][self.shadow_tree.nclasses()]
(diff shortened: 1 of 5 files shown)


dtreeviz/trees.py
```diff
@@ -131,7 +131,7 @@ def leaf_sizes(self,
 
 
     def ctree_leaf_distributions(self,
-                                 display_type: str = "plot",
+                                 display_type: ("plot", "text") = "plot",
```
parrt, line 134: I'm not sure I'm a big fan of plotting versus getting a string back (or printing). Seems like you would want something that plots and another function that returns Data structure indicating leaf id -> count, right?

tlapusan: yes, indeed. This was from beginning :)

the "text" version was my solution from that old time in case the plot contains too many leaves and hard to get the node id.

parrt: oh "text" is already in there and we just updated the type? ok.

tlapusan: yes, I just put there the possible options the user could choose.

PR #263, opened 2023-02-16
262 categorical features support for tf df

tlapusan:
should fix #262

PR #265, opened 2023-02-17
Tf catvars from tlapusan

parrt:
Starting from @tlapusan, raising exceptions when categorical variable's are used in feature space. I just don't know how to display that at the moment. Popping the version to 2.2.0 to incorporate categorical vars. 

Derived from https://github.com/parrt/dtreeviz/pull/263

We still need to test the decision notes more, such as when there are lots of levels in the catvars.


dtreeviz/trees.py
```diff
@@ -1591,6 +1623,18 @@ def _ctreeviz_bivar(shadow_tree, fontsize, ticks_fontsize, fontname, show,
     _format_axes(ax, shadow_tree.feature_names[featidx[0]], shadow_tree.feature_names[featidx[1]],
                  colors, fontsize, fontname, ticks_fontsize=ticks_fontsize, grid=False)
 
+    if 'splits' in show:
+        plt.draw()
+        print("labels", ax.get_xticklabels(), ax.get_yticklabels())
```
mepland, line 1628: @tlapusan is this print for debugging? Probably should be commented out.

mepland (changes requested):
The code looks good to me, besides one print statement. I don't see text axis tick labels in either of the demo notebooks though? `sex_label` is still 0 or 1.

dtreeviz/trees.py
parrt: sorry. yeah, debugging.

parrt:
> The code looks good to me, besides one print statement. I don't see text axis tick labels in either of the demo notebooks though? `sex_label` is still 0 or 1.

We have an updated the notebooks yet but I do have tutorial I'm working on that uses categoricals. Take a look at `testing/tf-catvars.py` for some playing around.

I need to fix `testing/tf-catvars2.py`... Working on that now hi cardinality categorical fails.

mepland (approved):

PR #277, opened 2023-03-06
Include string columns in node stats

### LINKED ISSUE #276: node_stats for nodes with features including string values.
When creating a data frame from a numpy, by default it converts all dtypes to object in case it contains different dtypes (float and str in our case)...
https://stackoverflow.com/questions/61346021/create-a-mixed-type-pandas-dataframe-using-an-numpy-array-of-type-object

Solution:

```
df = pd.DataFrame(self.shadow_tree.X_train, columns=self.shadow_tree.feature_names).convert_dtypes()
return df.iloc[node_samples[node_id]].describe(include='all')
```

tlapusan:
fix #276

### DIFF
--- a/dtreeviz/trees.py
+++ b/dtreeviz/trees.py
@@ -803,8 +803,8 @@ class DTreeVizAPI:
         """
 
         node_samples = self.shadow_tree.get_node_samples()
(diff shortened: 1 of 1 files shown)


tlapusan:
@parrt could you take a look at this ?

parrt:
Thanks!

PR #282, opened 2023-03-18
Make sklearn visualisations to support validation datasets.

tlapusan:


parrt:
Per my comment on the other PR. I'm not sure I understand the purpose :)

PR #292, opened 2023-04-28
Use prediction from shadow_tree instead of recalculating the mean

### LINKED ISSUE #291: _regr_leaf_viz calculates the mean for prediction value.
_regr_leaf_viz() function calculates itself the mean of values for leaf prediction (for regression tree)!

It should use the node.prediction() instead because in some cases we can take the leaf prediction value directly from the tree metadata, which is the correct way to do.

For example, the sklearn by default calculates the median for leaf prediction -> dtreeviz shows the mean, which is not correct.

tlapusan:
should close #291


tlapusan:
@parrt this one also :)

PR #328, opened 2024-08-22
Fix TF-DF version check.

rstz:
After TF-DF 1.10.0 was released, this check no longer worked. See https://www.tensorflow.org/decision_forests/tutorials/dtreeviz_colab

### DIFF
--- a/dtreeviz/models/tensorflow_decision_tree.py
+++ b/dtreeviz/models/tensorflow_decision_tree.py
@@ -56,7 +56,8 @@ class ShadowTensorflowTree(ShadowDecTree):
         return self.children_right
 
     def is_classifier(self) -> bool:
-        if tensorflow_decision_forests.__version__<'1.2.0':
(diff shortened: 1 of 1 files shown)


tlapusan:
Thanks @rstz, I will take a look soon.

tlapusan:
@rstz curious for this if condition "len(tfdf_version)<2". Looking at https://pypi.org/project/tensorflow-decision-forests/#history, it seems they have all 3 numbers.

If you could put spaces between operators would be great. Thanks.

Searching on internet how this could be otherwise solve, there seems to be this library https://packaging.pypa.io/en/stable/version.html, but I think would be to much to add it only for a simple check :)

rstz:
The if-statement is really just preventing an array out of bounds for later in that same line.

Another alternative to this PR would be to just remove support for the (quite outdated) TF-DF versions prior to 1.2.0 - happy to do this as well

tlapusan:
@parrt looks good for me to merge ! Thanks @rstz

